PR #9, opened 2018-12-03
Add property  for logFile location for launch and attach.

dschaefer:
This will be passed to the adapter to set the location of the log file.


dschaefer:
rerun tests

PR #12, opened 2018-12-19
Adjust to adapter output moving to dist.

dschaefer:


dschaefer:
Need master of the adapter to pick up the move to dist.

PR #14, opened 2019-01-09
 Add cdt-gdb-adapter to bundledDependencies

simark:
For some reason, when running "yarn pack", the out directory is
excluded.  For Theia, we would like to create archives that we can
download and use, and it works better if the "out" directory containing
the compiled scripts is included.  I suppose it won't change anything
for when it is used as a VSCode extension.

Signed-off-by: Simon Marchi <simon.marchi@ericsson.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,5 +137,8 @@
     "vscode": "^1.1.21",
     "webpack": "^4.25.1",
     "webpack-cli": "^3.1.2"
-  }
+  },
+  "bundledDependencies": [
+    "cdt-gdb-adapter"
+  ]
 }


simark:
Err, that's not relevant anymore, since we stopped using the submodule...

Specifying `node_modules/cdt-gdb-adapter` in `"files"` doesn't work, as it's automatically ignored... I'll have to figure out something.

simark:
Ok, found a way through `bundledDependencies`.  This is ready to be reviewed.

dschaefer:
run tests

dschaefer:
See, by doing the force push, I lost what you did the first time making this conversation a bit confusing. I'd rather see the history.

simark:
> See, by doing the force push, I lost what you did the first time making this conversation a bit confusing. I'd rather see the history.

Yeah... I really miss Gerrit for this.

PR #15, opened 2019-01-11
Use tsc and yarn to build.

dschaefer:


dschaefer:
Just come cleanup I've had in my workspace for a while. Aligns build with the adapter.

PR #18, opened 2019-02-05
Add configuration options to open a GDB console

marechal-p:
Feature will only work on UNIX systems, and fallback to the default way
of starting the Debug Adapter if not supported.

Signed-off-by: Paul Maréchal <paul.marechal@ericsson.com>

---

Add new launch/attach configuration option.


dschaefer:
rerun tests. I've added you to the whitelist so they should run automatically, I think...

marechal-p:
Thank you!

PR #20, opened 2019-02-11
Update to pull cdt-gdb-adapter from npm.

dschaefer:


dschaefer:
OK, dying to see if this works :)

PR #22, opened 2019-02-13
Start of memory browser.

dschaefer:
Just opens the webview and populates it with a simple React component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,9 +8,17 @@
     "vscode": "^1.26.0"
   },
   "activationEvents": [
-    "onDebug"
+    "onDebug",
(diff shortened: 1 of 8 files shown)


dschaefer:
This is just to get the ball rolling. We have approval now to use React. I'm looking at using Blueprint for the UI framework. It has a pretty powerful table.

dschaefer:
Yeah, this doesn't work BTW. The path to MemoryBrowser.js is wrong. And I need to webpack the client.

dschaefer:
Still searching for a good infinite scrolling table. Pushing this so I can experiment with a couple of things.

PR #23, opened 2019-02-19
Initial implementation of Memory Browser

dschaefer:
Based on Ericsson's work in Theia. Lots more to do to make
it look good.

Highlights: use of sass for css. Messaging between client webview
and server running in the extension.


marechal-p:
I have the same build errors as Jenkins when trying to build:
```
ERROR in /home/emaapur/projects/cdt-gdb-vscode/src/memory/client/MemoryBrowser.tsx
[tsl] ERROR in /home/emaapur/projects/cdt-gdb-vscode/src/memory/client/MemoryBrowser.tsx(226,56)
      TS2339: Property 'data' does not exist on type 'MemoryResponse'.

ERROR in /home/emaapur/projects/cdt-gdb-vscode/src/memory/client/MemoryBrowser.tsx
[tsl] ERROR in /home/emaapur/projects/cdt-gdb-vscode/src/memory/client/MemoryBrowser.tsx(227,52)
      TS2339: Property 'address' does not exist on type 'MemoryResponse'.
```

Anything special to do?

marechal-p:
I have the same build errors as Jenkins when building:
```
ERROR in .../cdt-gdb-vscode/src/memory/client/MemoryBrowser.tsx
[tsl] ERROR in .../cdt-gdb-vscode/src/memory/client/MemoryBrowser.tsx(226,56)
      TS2339: Property 'data' does not exist on type 'MemoryResponse'.

ERROR in .../cdt-gdb-vscode/src/memory/client/MemoryBrowser.tsx
[tsl] ERROR in .../cdt-gdb-vscode/src/memory/client/MemoryBrowser.tsx(227,52)
      TS2339: Property 'address' does not exist on type 'MemoryResponse'.
```

Anything special to do/setup?

dschaefer:
Ah, yes, there was a mistake in the adapter. I'll push that fix now.

dschaefer:
Created https://github.com/eclipse-cdt/cdt-gdb-adapter/pull/86

Issue is that the client only gets the body field of the response so need a type for it.

dschaefer:
Any objections t this?

simark:
I don't really have time to try it (it's my last day at $current_job, so lot's of things to do), but could you just show a screenshot?  I don't plan to have any objections, I would just like to see how it looks.

dschaefer:
It looks a lot like the Theia one :). I'll try and do a screenshot in a sec.

dschaefer:
memorybrowser

simark:
Ohh awesome.  I would really recommend using a monospace font, it would look better if everything was aligned.

Do the settings (bytes per row, bytes per group, endianness) and their effect make sense to you?

dschaefer:
> Ohh awesome. I would really recommend using a monospace font, it would look better if everything was aligned.

Yup, we'll do a bunch of cleanup like that. It's all default fonts for now. I just wanted to get people familiar with the plumbing we're using. I also need to hunt for a good mono-spaced font that I think will go well with the rest. I guess the one that the terminal uses would work. I have to see what vscode passes in through the CSS variables.

> Do the settings (bytes per row, bytes per group, endianness) and their effect make sense to you?

I think so. Certainly the size of the word (bytes per group) needs to be configurable. And endianness is a big one for me working with ARM a lot and am getting tired of byteswapping in my head :).

Not sure we can dynamically change the number of bytes in a row to fill the width of the window so having it a manual setting should be fine for now.

dschaefer:
BTW, I did make one change to the Theia code by having each "group" in it's own table cell. That helped a lot with alignment.

dschaefer:
OK, pushing now.

simark:
> I also need to hunt for a good mono-spaced font that I think will go well with the rest. I guess the one that the terminal uses would work. I have to see what vscode passes in through the CSS variables.

`font-family: monospace;` :)

> BTW, I did make one change to the Theia code by having each "group" in it's own table cell. That helped a lot with alignment.

Ok, we didn't see the need for it since the monospace font keeps everything aligned.  But at some point we might want to let the user select some cells, to edit values or something else.  So having them in separate table cells might help for that.

> Not sure we can dynamically change the number of bytes in a row to fill the width of the window so having it a manual setting should be fine for now.

I know that I wouldn't like the number of bytes per row to auto adjust with the width and end up with non "round" numbers (not sure how to say that in english).  For example, I wouldn't find it practical to end up with 29 bytes per row.  If it did auto adjust by only use multiples of 16 bytes per row, it might be ok.

dschaefer:
> > I also need to hunt for a good mono-spaced font that I think will go well with the rest. I guess the one that the terminal uses would work. I have to see what vscode passes in through the CSS variables.
> 
> `font-family: monospace;` :)

If it's good ;).

> 
> > BTW, I did make one change to the Theia code by having each "group" in it's own table cell. That helped a lot with alignment.
> 
> Ok, we didn't see the need for it since the monospace font keeps everything aligned. But at some point we might want to let the user select some cells, to edit values or something else. So having them in separate table cells might help for that.

Yeah, we might end up with a precanned React table widget to allow editing cells and such. And we're also looking at infinite scrolling. But that's down the road a bit.

> > Not sure we can dynamically change the number of bytes in a row to fill the width of the window so having it a manual setting should be fine for now.
> 
> I know that I wouldn't like the number of bytes per row to auto adjust with the width and end up with non "round" numbers (not sure how to say that in english). For example, I wouldn't find it practical to end up with 29 bytes per row. If it did auto adjust by only use multiples of 16 bytes per row, it might be ok.

Agreed.

simark:
> Yeah, we might end up with a precanned React table widget to allow editing cells and such. And we're also looking at infinite scrolling. But that's down the road a bit.

For infinite scrolling, I had two ideas:

- Keep a <table>, capture mousewheel events.  When the user goes up, just add X rows on top, remove X rows at the bottom.
- Replace the table with a canvas, also handle mouse events manually and update what is drawn accordingly.

dschaefer:
Clusterize is interesting. https://clusterize.js.org

I've also tried React virtualized, https://bvaughn.github.io/react-virtualized/#/components/Table

Though those aren't really infinite and relying on you setting a size. Mind you that size can be very big.

Mikhail K, also mentioned on the Gitter channel he was looking at this.

Never thought of using canvas. I wonder how well that would perform. Interesting approach though.

simark:
The xterm.js project (implements a terminal emulator, made for VSCode and used by Theia as well) uses a canvas.  They used to have it dom-based (lots of <div>s I suppose), but switched to canvas for performance reasons:

https://weihanglo.tw/posts/2017/how-is-new-terminal-in-vs-code-so-fast/

So surely performance wouldn't be a problem.  My concern would be that it's more efforts to implement.

simark:
Also, a little tangent: I don't think it would be really useful to have a scroll bar.  In theory, that scrollbar would have to represent the whole address space, so it wouldn't really be useful in practice.  Just some buttons to go up and down (equivalent to mousewheel up and down) would be fine, I think.

dschaefer:
The other issue with the browser scrollbar is that there is a maximum size to the height of a scrollable thing. I ran into that with a very large data set once. Scrolling with the scroll wheel starts to jump over rows.

PR #32, opened 2019-12-19
Fix syntax error in package.json

mkuznyetsov:
Fix the error in package.json, that would break the build of the project


jonahgraham:
Thank you @mkuznyetsov - sorry for the careless previous commit. BTW how did you catch it - I missed the red dot on https://ci.eclipse.org/cdt. Are you consuming this somewhere?

PR #49, opened 2021-06-01
Make it easier to extend the vscode extension

alicetrifu:


jonahgraham:
@paul-marechal Do you extend this code? Is what @alicetrifu proposing reasonable?

src/extension.ts
```diff
@@ -9,6 +9,7 @@
  *********************************************************************/
 import { ExtensionContext, commands, window } from 'vscode';
 import { MemoryServer } from './memory/server/MemoryServer';
+export { MemoryServer } from './memory/server/MemoryServer';
```
paul-marechal, line 12: Why are you exporting this from `extension.ts`?

alicetrifu: I've exported this in order to be extended. At first, I've only exported this in index.ts, but the class was not recognized as exported until this change was made.

paul-marechal: Ok found why: https://github.com/eclipse-cdt/cdt-gdb-vscode/blob/0e3779a19addae1c86ee3afb1514e420c6457d8c/package.json#L17

jonahgraham: @paul-marechal - does this mean you are ok with this part of the change? What about the rest of the change?

src/index.ts
```diff
@@ -0,0 +1,5 @@
+export {MemoryServer} from './memory/server/MemoryServer';
+export * from './memory/client/MemoryBrowser';
+export * from './memory/client/MessageBroker';
+export * from './memory/client/index'
+export * from './extension';
```
paul-marechal, line 5: Is this still useful to you @alicetrifu given what you mentioned about exporting from `src/extension.ts`?

src/extension.ts
paul-marechal: The changes looks ok, although I'm just curious about how @alicetrifu is actually extending this extension? It would help to see if there is a better approach or not.

jonahgraham: Thanks @paul-marechal - I look forward to @alicetrifu answering some of these questions.

alicetrifu: We need to create a vsix file specific to us. It will use the cdt-gdb-adpater extension,  as well, which will also contain, additionally, commands specific to the targets we using. By extending this, we will be able to have an extension that will contain specific information about the debugging process that users will be able to do on our supported targets.

src/index.ts
alicetrifu: It is. I need the classes from "./memory/client", in order to properly work. Regarding the src/extension.ts , I needed for some reason to do the export here as well, because the class was not recognized as exported until was on both places.

paul-marechal: How do you consume this package?

If you consume it as a regular npm package, then exporting from `src/extension.ts` is enough as it is the exposed script when doing `require('cdt-gdb-vscode')`.

This `src/index.ts` file is only ever useful if you do something like `require('cdt-gdb-vscode/out')`.

Just trying to straighten the setup here because it looks odd to export the same things from two separate places.

alicetrifu: @paul-marechal indeed, you are right. I've edited my extension and I do not need the src/index.ts. I'm doing the import by 'cdt-gdb-vscode'. I will do the push again without this file and as well the changes from tsconfig,json

src/extension.ts
jonahgraham: Thanks @alicetrifu - this appears to be fully resolved now. @paul-marechal if you have any further feedback we can follow up in a subsequent issue/pr.

jonahgraham (approved):

PR #52, opened 2021-08-18
Upgrade cdt gdb adapter to 0.15.0 and yarn upgrade

jonahgraham:


jonahgraham:
Not sure if this should be 0.15.0 or updated to 0.16.0-next.

jonahgraham:
For now that is the same content (except for the version number)

PR #68, opened 2022-04-29
fix issue in lengthReq

quyettrinhrvc:
Hi @jonahgraham ,
I think this is an issue with data in "Length" box of Memory Browser.
Currently, the size of memory to be viewed is the value of "Location" box (address) but not "Length" box.
Could you please help to check it?


jonahgraham:
Thanks you @quyettrinhrvc!

PR #76, opened 2022-11-29
Support Memory Browser for multiple child

trongle0504:
Support Memory Browser for multiple children

### DIFF
--- a/src/memory/client/MemoryBrowser.tsx
+++ b/src/memory/client/MemoryBrowser.tsx
@@ -77,11 +77,14 @@ interface State {
   bytesPerRow: number;
   bytesPerGroup: number;
   endianness: 'le' | 'be';
+  childNames: { id: string; name: string }[];
 }
 
 export class MemoryBrowser extends React.Component<Props, State> {
   private addressReq = '';
   private lengthReq = '512';
+  private childReq = '0';
+  public static firtTime: boolean = true;
 
   constructor(props: Props) {
     super(props);
@@ -89,7 +92,35 @@ export class MemoryBrowser extends React.Component<Props, State> {
       bytesPerRow: 32,
       bytesPerGroup: 8,
       endianness: 'le',
+      childNames: [],
     };
+  }
+
+  async getAvailableChilds() {
+    if (this.state.childNames.length === 0) {
+      try {
+        const result = await messageBroker.send({
+          command: 'ReadMemory',
+          args: {
+            address: this.addressReq,
+            length: parseInt(this.lengthReq),
+            child: this.childReq,
+          },
+        });
+        let childArray: {
+          id: any;
+          name: any;
+        }[] = [];
+        result.result?.child?.map(function (val: any, index: any) {
+          childArray.push({ id: index, name: val });
+        });
+        this.setState({
+          childNames: childArray,
+        });
+      } catch (err) {
+        this.setState({ error: <h3>{err + ''}</h3> });
+      }
+    }
   }
 
   async sendReadMemoryRequest() {
@@ -106,6 +137,7 @@ export class MemoryBrowser extends React.Component<Props, State> {
           args: {
             address: this.addressReq,
             length: parseInt(this.lengthReq),
+            child: this.childReq,
           },
         });
         this.setState({ memory: result.result });
@@ -145,6 +177,7 @@ export class MemoryBrowser extends React.Component<Props, State> {
             onChange={(event) => (this.lengthReq = event.target.value)}
           />
         </div>
+        {this.renderChildName()}
         <div className="input-group">
           <button onClick={() => this.sendReadMemoryRequest()}>Go</button>
         </div>
@@ -273,11 +306,17 @@ export class MemoryBrowser extends React.Component<Props, State> {
 
       rows.push(
         <tr key={rowOffset}>
-          <td className='monofont' key={`addr${rowOffset}`}>{addressStr}</td>
+          <td className="monofont" key={`addr${rowOffset}`}>
+            {addressStr}
+          </td>
           {data.map((group, index) => (
-            <td className='monofont' key={`data${rowOffset},${index}`}>{group}</td>
+            <td className="monofont" key={`data${rowOffset},${index}`}>
+              {group}
+            </td>
           ))}
-          <td className='monofont' key={`asc${rowOffset}`}>{asciiStr}</td>
+          <td className="monofont" key={`asc${rowOffset}`}>
+            {asciiStr}
+          </td>
         </tr>
       );
     }
@@ -317,5 +356,37 @@ export class MemoryBrowser extends React.Component<Props, State> {
         {this.renderMemory()}
       </div>
     );
+  }
+
+  renderChildName() {
+    if (MemoryBrowser.firtTime) {
+      this.getAvailableChilds();
+      MemoryBrowser.firtTime = false;
+    }
+    const { childNames } = this.state;
+    let childNamesList =
+      childNames.length > 0 &&
+      childNames.map((item, i) => {
+        return (
+          <option key={i} value={item.id}>
+            {item.name}
+          </option>
+        );
+      }, this);
+    if (childNames.length > 0) {
+      return (
+        <div className="input-group">
+          <label>Child</label>
+          <select
+            defaultValue={this.childReq}
+            onChange={(event) => (this.childReq = event.target.value)}
+          >
+            {childNamesList}
+          </select>
+        </div>
+      );
+    } else {
+      return <p></p>;
+    }
   }
 }


src/memory/client/MemoryBrowser.tsx
```diff
@@ -106,6 +137,7 @@ export class MemoryBrowser extends React.Component<Props, State> {
           args: {
             address: this.addressReq,
             length: parseInt(this.lengthReq),
+            child: this.childReq,
```
jonahgraham, line 140: It doesn't hurt to include this when sending to cdt-gdb-adapter/Memory, but continue to ensure that it isn't necessary to have so that the memory code works directly with cdt-gdb-adapter.

src/memory/client/MemoryBrowser.tsx
```diff
@@ -77,21 +77,52 @@ interface State {
   bytesPerRow: number;
   bytesPerGroup: number;
   endianness: 'le' | 'be';
+  childNames: { id: string; name: string }[];
```
jonahgraham, line 80: Can you change `childNames` to `childrenNames` and similar changes elsewhere in the code.

src/memory/client/MemoryBrowser.tsx
```diff
@@ -318,4 +357,36 @@ export class MemoryBrowser extends React.Component<Props, State> {
...
+          </option>
+        );
+      }, this);
+    if (childNames.length > 0) {
+      return (
+        <div className="input-group">
+          <label>Child</label>
+          <select
+            defaultValue={this.childReq}
+            onChange={(event) => (this.childReq = event.target.value)}
```
jonahgraham, line 382: Please check the types here. As mentioned in the cdt-amalgamator change indexes should be numbers.

src/memory/client/MemoryBrowser.tsx
```diff
@@ -77,21 +77,52 @@ interface State {
   bytesPerRow: number;
   bytesPerGroup: number;
   endianness: 'le' | 'be';
+  childNames: { id: string; name: string }[];
 }
 
 export class MemoryBrowser extends React.Component<Props, State> {
   private addressReq = '';
   private lengthReq = '512';
+  private childReq = '0';
```
jonahgraham, line 86: Please check the types here. As mentioned in the cdt-amalgamator change indexes should be numbers.

src/memory/client/MemoryBrowser.tsx
```diff
@@ -318,4 +357,36 @@ export class MemoryBrowser extends React.Component<Props, State> {
       </div>
     );
   }
+
+  renderChildName() {
+    if (MemoryBrowser.firtTime) {
+      this.getAvailableChilds();
+      MemoryBrowser.firtTime = false;
```
jonahgraham, line 364: This needs to be reset to true so that future debug sessions fetch the availableChildren again?

src/memory/client/MemoryBrowser.tsx
```diff
@@ -77,21 +77,52 @@ interface State {
...
+            address: this.addressReq,
+            length: parseInt(this.lengthReq),
+            child: this.childReq,
+          },
+        });
+        let childArray: {
+          id: any;
+          name: any;
+        }[] = [];
+        result.result?.child?.map(function (val: any, index: any) {
```
jonahgraham, line 114: Don't use any if at all possible.

jonahgraham, line 114: This isn't how to use map, map should transform one thing to another, e.g.:

```typescript
const childrenArray = result.result?.child?.map((val, index) => {id: index, name: val});
```

Please see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map for a comprehensive overview of map in javascript.

jonahgraham (changes requested):
This looks very promising to me. The basic idea looks correct, but I am concerned about three things:

- Handling of types, this is the use of `any` and index values stored as `string`
- How the available children get refreshed/updated/validated
- How this works without cdt-amalgamator

src/memory/client/MemoryBrowser.tsx
trongle0504: I have updated it

src/memory/client/MemoryBrowser.tsx
trongle0504: I have updated it

src/memory/client/MemoryBrowser.tsx
trongle0504: I have updated it

src/memory/client/MemoryBrowser.tsx
trongle0504: I have updated it

src/memory/client/MemoryBrowser.tsx
trongle0504: I just want to `getAvailableChilds` function is called only one time when Memory Browser view is opened

src/memory/client/MemoryBrowser.tsx
trongle0504: I have defined the type for all of them.

trongle0504: The input of this function is an array and the output is an array of objects for example:
input: [RX, RL, RZ]
output: [{id: 0, name: RX}, {id: 1, name: RL}, {id: 2, name: RZ}]

src/memory/client/MemoryBrowser.tsx
```diff
@@ -77,21 +77,50 @@ interface State {
...
     super(props);
     this.state = {
       bytesPerRow: 32,
       bytesPerGroup: 8,
       endianness: 'le',
+      childrenNames: [],
     };
   }
 
+  async getAvailableChilds() {
```
jonahgraham, line 99: Can you rename this `getAvailableChildren`?

src/memory/client/MemoryBrowser.tsx
```diff
@@ -77,21 +77,50 @@ interface State {
...
+    try {
+      const result = await messageBroker.sendGetChildrenNames({
+        command: 'getChildDapNames',
+        args: {
+          child: this.childReq,
```
jonahgraham, line 104: This argument does not make sense to me. What does a childReq mean in the context of getting all the children?

src/memory/client/MemoryBrowser.tsx
jonahgraham: Yes, I understand that, but map is normally used to map from one type to another. I gave you the standard way of writing that code, but you didn't change it, not explain what your issue is.

src/memory/client/MemoryBrowser.tsx
```diff
@@ -77,21 +77,50 @@ interface State {
...
+      let childArray: {
+        id: number;
+        name: string;
+      }[] = [];
+      result.result?.child?.map(function (val: string, index: number) {
+        childArray.push({ id: index, name: val });
+      });
+      this.setState({
+        childrenNames: childArray,
+      });
```
jonahgraham, line 116: This isn't how to use map, map should transform one thing to another, e.g.:

Please see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map for a comprehensive overview of map in javascript.

```suggestion
      const childrenNames = result.result?.child?.map((val, index) => {id: index, name: val});
      this.setState({
        childrenNames,
      });
```

(you may want `|| []` at the end of the first line if you don't want childrenNames to be undefined.

src/memory/client/MemoryBrowser.tsx
```diff
@@ -101,14 +130,26 @@ export class MemoryBrowser extends React.Component<Props, State> {
...
+        } else {
+          const result = await messageBroker.send({
+            command: 'ReadMemory',
+            args: {
+              address: this.addressReq,
+              length: parseInt(this.lengthReq),
+            },
+          });
+          this.setState({ memory: result.result });
+        }
```
jonahgraham, line 152:
```suggestion
        const result = await messageBroker.send({
          command: 'ReadMemory',
          args: {
            address: this.addressReq,
            length: parseInt(this.lengthReq),
            child: this.state.childrenNames.length ? this.childReq : undefined,
          },
        });
        this.setState({ memory: result.result });
```

src/memory/client/MessageBroker.ts
```diff
@@ -7,21 +7,37 @@
...
+    GetChildDapNames,
+} from '../common/messages';
 
 declare function acquireVsCodeApi(): any;
 const vscode = acquireVsCodeApi();
 
 class MessageBroker {
     private currentToken = 1;
-    private queue: { [token: number]: (result: ServerResponse) => void } = {};
+    private queue1: { [token: number]: (result: ServerResponse) => void } = {};
```
jonahgraham, line 23: I don't understand why two queues now. Can you please explain it? It gets extra confusing to me because below you have two different event listeners installed on the same event too.

jonahgraham (changes requested):

src/memory/client/MemoryBrowser.tsx
trongle0504: I have updated it.

src/memory/client/MemoryBrowser.tsx
trongle0504: I deleted it. Because we just need to post a comment cdt-amalgamator/getChildDapNames and get a response. It is not necessary to enclose args.

src/memory/client/MemoryBrowser.tsx
trongle0504: I have updated it.

src/memory/client/MemoryBrowser.tsx
trongle0504: I have updated it.

package.json
```diff
@@ -486,6 +486,7 @@
   },
   "dependencies": {
     "cdt-gdb-adapter": "^0.0.16-next",
+    "cdt-amalgamator": "^0.0.5",
```
jonahgraham, line 489: Please ensure changes to dependencies have corresponding changes to yarn.lock.

src/memory/client/MemoryBrowser.tsx
jonahgraham: Resolving here, but I don't see the change. I will make a new comment in the updated code.

src/memory/client/MemoryBrowser.tsx
```diff
@@ -101,14 +124,28 @@ export class MemoryBrowser extends React.Component<Props, State> {
...
         });
-        const result = await messageBroker.send({
-          command: 'ReadMemory',
-          args: {
-            address: this.addressReq,
-            length: parseInt(this.lengthReq),
-          },
-        });
-        this.setState({ memory: result.result });
+        if (this.state.childrenNames.length > 0) {
```
jonahgraham, line 127:
```suggestion
```

src/memory/client/MemoryBrowser.tsx
```diff
@@ -101,14 +124,28 @@ export class MemoryBrowser extends React.Component<Props, State> {
...
+        } else {
+          const result = await messageBroker.send({
+            command: 'ReadMemory',
+            args: {
+              address: this.addressReq,
+              length: parseInt(this.lengthReq),
+            },
+          });
+          this.setState({ memory: result.result });
+        }
```
jonahgraham, line 148:
```suggestion
```

src/memory/client/MemoryBrowser.tsx
```diff
@@ -101,14 +124,28 @@ export class MemoryBrowser extends React.Component<Props, State> {
...
-        this.setState({ memory: result.result });
+        if (this.state.childrenNames.length > 0) {
+          const result = await messageBroker.send({
+            command: 'ReadMemory',
+            args: {
+              address: this.addressReq,
+              length: parseInt(this.lengthReq),
+              child: this.state.childrenNames.length
+                ? this.childReq
+                : undefined,
```
jonahgraham, line 135: You almost collapsed the two calls to `ReadMemory` correctly. But in this update, the else of this `?:` expression is never used because at line 127 you only enter this if block if length is > 0

src/memory/server/MemoryServer.ts
```diff
@@ -99,9 +110,44 @@ export class MemoryServer {
...
+                    });
+                } else {
+                    const result: MemoryContents = await session.customRequest(
+                        'cdt-amalgamator/Memory',
+                        request.args
+                    );
+                    this.sendResponse(panel, request, {
+                        result,
+                    });
+                }
```
jonahgraham, line 129: I don't think we need the whole thing copy-pasted, just the one parameter is different.
```suggestion
                    const result: MemoryContents = await session.customRequest(
                        request.args.child === undefined ? 'cdt-gdb-adapter/Memory' : 'cdt-amalgamator/Memory',
                        request.args
                    );
                    this.sendResponse(panel, request, {
                        result,
                    });
```

jonahgraham (changes requested):

package.json
jonahgraham: The 0.0.6 version has been published that has these changes: https://www.npmjs.com/package/cdt-amalgamator/v/0.0.6

src/memory/server/MemoryServer.ts
trongle0504: I have updated it.

src/memory/client/MemoryBrowser.tsx
trongle0504: Sorry, my mistake. I have updated it.

src/memory/client/MemoryBrowser.tsx
trongle0504: I have updated it.

package.json
trongle0504: I have updated the current version `0.0.6`. And added yarn.lock file

trongle0504:
Hi @jonahgraham ,
I have updated code.
Could you please help me review it?

Thank you!

jonahgraham (approved):

PR #85, opened 2023-03-02
Updating the types/vscode version

trongle0504:
We need to update the new version for types/vscode.


trongle0504:
Hi @jonahgraham,
We need to update the new version for `types/vscode`.
Additionally, could you update the latest version for the **cdt-amalgamator** and **cdt-gdb-adapter** repositories?
Currently, we have added some new code to these two repositories, but the new version has not been updated for users yet.

Thank you!

jonahgraham:
I updated all the versions in https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/commit/4cbccae26ed042d6e37a1970b334a8b8fd7f2472

jonahgraham:
>  Additionally, could you update the latest version for the **cdt-amalgamator** and **cdt-gdb-adapter** repositories? 

I did those ones too.

PR #89, opened 2023-04-28
Support disable/enable hardware breakpoint

trongle0504:
Currently, enabling hardwareBreakpoint in launch.json argument allows the debugger to set hardware break-point or software break-point only ("hardwareBreakpoint": true). We need to support enabling/disabling setting SW and HW break-point simultaneously while debugging.


trongle0504:
Hi @jonahgraham -san,

Could you please help me review the code?
Please see the document attached to ticket **IDE-63527** for details.

I also commit code in repository cdt-gdb-adapter.
https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/264 

Thank you!

src/breakpoint/DisableHWBreakpoint.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    private readonly registerCommands = (context: ExtensionContext) => {
+        context.subscriptions.push(
+            vscode.commands.registerCommand(
+                'cdt.debug.disableHWBreakpoint',
+                async () => {
+                    const session = vscode.debug.activeDebugSession;
+                    if (session) {
+                        await vscode.commands.executeCommand(
```
jonahgraham, line 16: Can you simplify this to a single context?

src/breakpoint/DisableHWBreakpoint.ts
```diff
@@ -0,0 +1,34 @@
...
+                            'setContext',
+                            'disableHardware',
+                            false
+                        );
+                        await vscode.commands.executeCommand(
+                            'setContext',
+                            'enableHardware',
+                            true
+                        );
+                        await session.customRequest(
```
jonahgraham, line 26: Please pass the desired state to the adapter.

package.json
```diff
@@ -503,6 +513,18 @@
           "group": "navigation",
           "when": "debugType == amalgamator"
         }
+      ],
+      "editor/lineNumber/context": [
+        {
+          "command": "cdt.debug.enableHWBreakpoint",
+          "group": "navigation",
+          "when": "(!enableHardware && !disableHardware) || enableHardware"
```
jonahgraham, line 521: Having `enableHardware` twice in this expression is unexpected.

jonahgraham (changes requested):
As mentioned in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/264 using a toggle for the new command is problematic as the initial state of hw breakpoint as set in launch.json is not taken into account here and leaves the UI commands in an inconsistent state.

I have serious concerns about the overall usability of this design and I hope that once you share the information in the below mentioned document I can understand the workflows. 

For example, are you supposed to be able to change hw -> sw (or back) in the middle of a debug session?

> Please see the document attached to ticket **IDE-63527** for details.

Please provide required information in the open - this is an open source project so should not require access to private information to understand this.

trongle0504:
> As mentioned in [eclipse-cdt-cloud/cdt-gdb-adapter#264](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/264) using a toggle for the new command is problematic as the initial state of hw breakpoint as set in launch.json is not taken into account here and leaves the UI commands in an inconsistent state.
> 
> I have serious concerns about the overall usability of this design and I hope that once you share the information in the below mentioned document I can understand the workflows.
> 
> For example, are you supposed to be able to change hw -> sw (or back) in the middle of a debug session?
> 
> > Please see the document attached to ticket **IDE-63527** for details.
> 
> Please provide required information in the open - this is an open source project so should not require access to private information to understand this.

Hi @jonahgraham ,

I have some changes in the purpose of this PR compared to the original one. 

Below is the description of this PR:

The purpose of this PR is to support **toggle breakpoint** being able to change the **Hardware breakpoint -> Software breakpoint (or back)** in the middle of a debug session.

**Case 1:** If `hardwareBreakpoint` is not configured in launch.json, the default setting will be SW breakpoint.
1. After launching the debugger, the user right-clicks on the lineNumber and selects **Toggle Breakpoint**.
Screenshot_24

2. The **Debug Console** will display a message indicating that HW breakpoint is now in use.
Screenshot_25
3. The user can continue to use Toggle Breakpoint to switch between HW and SW breakpoints.

**Case 2:** If `"hardwareBreakpoint": true` is configured in launch.json, the default setting will be HW breakpoint.
1. After launching the debugger, the user right-clicks on the lineNumber and selects **Toggle Breakpoint**.
Screenshot_24

2. The **Debug Console** will display a message indicating that the SW breakpoint is now in use.
Screenshot_27

3. The user can continue to use Toggle Breakpoint to switch between HW and SW breakpoints.

Could you please help me review it again?

Thank you!

src/breakpoint/DisableHWBreakpoint.ts
trongle0504: The solution has been changed to use `Toggle Breakpoint`

package.json
trongle0504: The solution has been changed to use `Toggle Breakpoint`

src/breakpoint/DisableHWBreakpoint.ts
trongle0504: Since using `Toggle Breakpoint`, I think there is no need to send status in `customRequest`

package.json
```diff
@@ -16,13 +16,14 @@
   "homepage": "https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode#readme",
   "main": "./out/extension",
   "engines": {
-    "vscode": "^1.76.0"
+    "vscode": "^1.78.0"
```
jonahgraham, line 19: Normally I would want version change like this to be split out - but since there is a dependency here on https://github.com/microsoft/vscode/issues/175945 I think it is ok this time.

jonahgraham (changes requested):
> The purpose of this PR is to support **toggle breakpoint** being able to change the **Hardware breakpoint -> Software breakpoint (or back)** in the middle of a debug session.

As you have implemented it now the toggle will apply to future breakpoints that are inserted by GDB - that intention is not obvious to the user. Was there an intention to change existing breakpoints from hw -> sw (or vice versa)?

Messages in the Debug Console are somewhat useful, but they can also be lost among lots of other output from the backend and it isn't visible if the Debug Console is not open.

VS Code has standard ways of notifying users. Please look at [showInformationMessage](https://code.visualstudio.com/api/references/vscode-api#window) and review [UX guidelines on notifications](https://code.visualstudio.com/api/ux-guidelines/notifications).

However in this case there is another problem. The current implementation provides no way to show user the current setting, leading to the user having to toggle the setting twice and looking at Debug Console to know the current setting.

Perhaps using the [status bar](https://code.visualstudio.com/api/ux-guidelines/status-bar) to show current state is useful? Or a checkmark next to a menu item?

The final problem I have with this solution is that there is no way to show users for already inserted breakpoints whether it is a hardware or software breakpoint - other than user typing `>info break` or similar in the CLI.

Note that I can accept a solution that does not address every UX items I have described above as some of them are really hard and may require changes to VSCode too. However the current solution is simply too far away from an intuitive design for users.

package.json
```diff
@@ -46,6 +47,10 @@
           "light": "images/SuspendAll_Light.png",
           "dark": "images/SuspendAll_Dark.png"
         }
+      },
+      {
+        "command": "cdt.debug.toggleBreakpoint",
+        "title": "Toggle Breakpoint"
```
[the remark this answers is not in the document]
trongle0504: I updated name to "Use Hardware breakpoints"

src/breakpoint/ToggleBreakpoint.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    private readonly registerCommands = (context: ExtensionContext) => {
+        context.subscriptions.push(
+            vscode.commands.registerCommand(
+                'cdt.debug.toggleBreakpoint',
+                async () => {
+                    const session = vscode.debug.activeDebugSession;
+                    if (session) {
+                        await session.customRequest(
+                            'cdt-gdb-adapter/toggleBreakpoint'
```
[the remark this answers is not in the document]
trongle0504: I created customRequest() to get the current state, then change the state and send it to set HW Breakpoint.
Screenshot_36

trongle0504:
> Messages in the Debug Console are somewhat useful, but they can also be lost among lots of other output from the backend and it isn't visible if the Debug Console is not open.
> 
> VS Code has standard ways of notifying users. Please look at [showInformationMessage](https://code.visualstudio.com/api/references/vscode-api#window) and review [UX guidelines on notifications](https://code.visualstudio.com/api/ux-guidelines/notifications).

I replaced the message at the Debug console to the one using VSCode's Notifications.
Screenshot_37
Screenshot_39

trongle0504:
Hi @jonahgraham ,
I've updated some of the changes you suggested including:
- Changed notify when setting HW breakpoint for users by standard ways of notifying users of VSCode
- Show current status of the HW breakpoint to user at status bar.
- Change the name "Toggle breakpoints" to the name "Use Hardware breakpoints".
- Use `cdt-gdb-adapter/getHWBreakpoint`, `cdt-gdb-adapter/setHWBreakpoint` to get and set HW state breakpoint to GDB using `customRequest()`

Could you please help me review it again?

Thank you!

package.json
jonahgraham: Is this version bump still needed since you aren't using new feature? Lets split out the version bump to a new PR.

src/breakpoint/BreakpointStatusBarItem.ts
```diff
@@ -0,0 +1,23 @@
+import { window, StatusBarAlignment, StatusBarItem } from 'vscode';
```
jonahgraham, line 1:
```suggestion
/*********************************************************************
 * Copyright (c) 2023 Renesas Electronics Corporation and others
 *
 * This program and the accompanying materials are made
 * available under the terms of the Eclipse Public License 2.0
 * which is available at https://www.eclipse.org/legal/epl-2.0/
 *
 * SPDX-License-Identifier: EPL-2.0
 *********************************************************************/
import { window, StatusBarAlignment, StatusBarItem } from 'vscode';
```

src/breakpoint/SetHWBreakpoint.ts
```diff
@@ -0,0 +1,78 @@
+import { ExtensionContext, window } from 'vscode';
```
jonahgraham, line 1:
```suggestion
/*********************************************************************
 * Copyright (c) 2023 Renesas Electronics Corporation and others
 *
 * This program and the accompanying materials are made
 * available under the terms of the Eclipse Public License 2.0
 * which is available at https://www.eclipse.org/legal/epl-2.0/
 *
 * SPDX-License-Identifier: EPL-2.0
 *********************************************************************/
import { ExtensionContext, window } from 'vscode';
```

src/breakpoint/BreakpointStatusBarItem.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export class BreakpointStatusBarItem {
```
jonahgraham, line 3: What is the purpose of this class? Why not use createStatusBarItem directly?

src/breakpoint/SetHWBreakpoint.ts
```diff
@@ -0,0 +1,78 @@
...
+                    '.vscode',
+                    'launch.json'
+                );
+                const config: vscode.WorkspaceConfiguration =
+                    vscode.workspace.getConfiguration(
+                        'launch',
+                        vscode.Uri.file(configPath)
+                    );
+                const HWBreakpointParameter: boolean | undefined =
+                    config?.configurations[0]?.hardwareBreakpoint;
```
jonahgraham, line 31: Is it possible to  get the current status from the backend rather than trying to maintain duplicate logic in the frontend?

jonahgraham (changes requested):
Some of the language needs some additional thought - i.e differentiating between action and state. We can refine that once we agree on the high level UX issues like below.

> > The final problem I have with this solution is that there is no way to show users for already inserted breakpoints whether it is a hardware or software breakpoint - other than user typing `>info break` or similar in the CLI.
> 
> Regarding this issue, I have investigated but I have no solution for this. Any ideas for this issue? or we will improve it in the future.

I need some more time to contemplate this aspect. It is related to other issues like preserving settings and setting expectations for users. Because there is no persistence and no way of identifying which type of breakpoint is used I fear we are giving users a false expectation. 

Consider for example this flow:

0. User sets use hw breakpoints in launch config
1. User inserts some hardware breakpoints
2. User starts debug session
3. User toggles to use software breakpoints
4. User inserts their software breakpoints
5. User presses "Toggle Active Breakpoints" (twice)
6. Now all breakpoints are software breakpoints

A similar situation:

0. User sets use hw breakpoints in launch config
1. User inserts some hardware breakpoints
2. User starts debug session
3. User toggles to use software breakpoints
4. User inserts their software breakpoints
5. User restarts their debug sessions
6. Now all breakpoints are hardware breakpoints (and likely many are not successfully inserted due to limitation on number of hardware breakpoints)

trongle0504:
I'm sorry, I would like to close this request because there are some complex cases to cover, and it may cause inconvenience for users. Thank you for taking the time to review the code.

Issue #95, opened 2023-08-04
cdt-gdb-vscode 0.0.101 not ending up on Open VSX


jonahgraham:
I pushed 0.0.101 to cdt-gdb-vscode a few days ago but it hasn't shown up on open vsx yet. I am not sure how to give this a kick or who to report this to.

@sdirix, @thegecko, @planger do any of you know how/who to ask to make latest cdt-gdb-vscode show up?

thegecko:
There have been some issues recently, e.g. 

https://github.com/eclipse/openvsx/issues/779

Recommend opening an issue there.

PR #96, opened 2023-08-30
Proposal: Package Extension

colin-grant-work:
At present, packaging this extension results in much of its `node_modules` getting packaged as well: `vsce ls` lists almost 600 files, including such things as the README for `node-gyp-build`. This PR packages the extension using ESBuild and modifies the `.vscodeignore` to use an include-list style.

Oddities and queries:
 - Why ESBuild and not Webpack? The same goals could be achieved with a more elaborate Webpack config rather than a switch to ESBuild. However, ESBuild is quite a bit faster, and (for now) easier to configure, and comes with fewer transitive dependencies. Even serializing TSC and ESBuild as I describe below, the build is still faster (caveats: on my machine, etc.) than the current parallel TSC + Webpack build, even though it's doing three bundling jobs (extension, debug adapter x2) that the current build doesn't do at all. On the other hand, it doesn't come with all of Webpack's bells and whistles, to wit:
 - ESBuild does no type checking, and the extension currently has no CI, so instead of using ESBuild directly, I've set the scripts to do TypeScript transpilation into the `out` directory and then bundle that with ESBuild into `dist`. There are alternatives to this:
     - A CI-oriented script that would do type checking, apart from the normal build process, while the build process uses ESBuild alone.
     - Using Webpack instead of ESBuild, as it combines bundling and typechecking transpilation. 
 - This PR applies ESBuild to the output of the CDT-GDB Adapter's compilation; alternatively, that package could ship a minified version, and this extension could consume that.
 - This extension depends on cdt-gdb-amalgamator and refers to a debug type `amalgamator` but doesn't declare such a debug type itself, so I haven't done anything to include that code separately. Is that correct, or should that code get special handling similar to that of the CDT GDB Adapter?
 - Currently, this extension includes source maps in its production releases. This code doesn't, but it would be easy to do so. `.d.ts` files would be slightly more work, but doable. I'm not sure whom those serve, though, since this extension isn't deployed as an NPM extension (as far as I know).


jonahgraham (approved):
I have no problem with this change - I wasn't active on this project when the current decisions were made and I have not examined any of this. 

PS Was ccd5af4451397f67889e4e4441a2a1b6a236f592 supposed to be part of this PR?

colin-grant-work:
> PS Was [ccd5af4](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/commit/ccd5af4451397f67889e4e4441a2a1b6a236f592) supposed to be part of this PR?

Yikes. Yes it was, as was the preceding commit. I'm not sure how I ended up pushing to `main` here. Apologies. I'll make a reversion MR and include them here instead. Perhaps we should protect the `main` branch from my fat fingers, as well?

jonahgraham:
> > PS Was [ccd5af4](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/commit/ccd5af4451397f67889e4e4441a2a1b6a236f592) supposed to be part of this PR?
> 
> Yikes. Yes it was, as was the preceding commit. I'm not sure how I ended up pushing to `main` here. Apologies. I'll make a reversion MR and include them here instead. Perhaps we should protect the `main` branch from my fat fingers, as well?

If there is a way to protect the main branch, but still allow review-less PRs to be merged please go ahead - the way to do it is to make a PR on this repo. AFAIU the main branch can only be protected if we require reviews, which, at least for releng work, is a blocker.

jonahgraham:
> PR on this repo

forgot link: https://github.com/eclipse-cdt-cloud/.eclipsefdn

jonahgraham:
Before this change gets pushed I am going to publish 0.0.102 version that picks up https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/289 for @parisa-mchp

jonahgraham:
@colin-grant-work thanks for the quick turnaround. If possible can you wait until the 0.0.102 version ends up on [openvsx](https://open-vsx.org/extension/eclipse-cdt/cdt-gdb-vscode) before merging this? With any luck that will be tomorrow and we won't have a recurrence of #95

colin-grant-work:
Yeah, no problem. I was expecting a lot more discussion, so I wasn't expecting to get this in immediately in any case!

colin-grant-work:
I've had some more time to confirm that this works downstream, if you're still alright with its merge, @jonahgraham?

colin-grant-work:
@jonahgraham, saw from the threads in the adapter that you'd queued up another publication of the plugin side. Shall I wait to merge this until tomorrow, in that case?

jonahgraham:
> @jonahgraham, saw from the threads in the adapter that you'd queued up another publication of the plugin side. Shall I wait to merge this until tomorrow, in that case?

Good idea. I think @xingmicrochip has a release coming up soon and want to pick up that fix, so it may be less disruptive to them to give them a day to get a version without this change in place. Thanks Colin for considering that.

asimgunes:
Hi, 

I like to provide some feedback related to this update. After this update, I got unable to run the `yarn build` command successfully in this project and I was getting the error show below:

```
$ node esbuild.js
X [ERROR] Could not resolve "../../build/Release/pty.node"

    node_modules/cdt-gdb-adapter/dist/native/pty.js:21:28:
      21 │         const pty = require('../../build/Release/pty.node');
         ╵                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

C:\Workspace\cdt-gdb-vscode\node_modules\esbuild\lib\main.js:1649
  let error = new Error(text);
              ^

Error: Build failed with 1 error:
node_modules/cdt-gdb-adapter/dist/native/pty.js:21:28: ERROR: Could not resolve "../../build/Release/pty.node"
    at failureErrorWithLog (C:\Workspace\cdt-gdb-vscode\node_modules\esbuild\lib\main.js:1649:15)
    at C:\Workspace\cdt-gdb-vscode\node_modules\esbuild\lib\main.js:1058:25
    at C:\Workspace\cdt-gdb-vscode\node_modules\esbuild\lib\main.js:1003:52
    at buildResponseToResult (C:\Workspace\cdt-gdb-vscode\node_modules\esbuild\lib\main.js:1056:7)
    at C:\Workspace\cdt-gdb-vscode\node_modules\esbuild\lib\main.js:1085:16
    at responseCallbacks.<computed> (C:\Workspace\cdt-gdb-vscode\node_modules\esbuild\lib\main.js:703:9)
    at handleIncomingPacket (C:\Workspace\cdt-gdb-vscode\node_modules\esbuild\lib\main.js:762:9)
    at Socket.readFromStdout (C:\Workspace\cdt-gdb-vscode\node_modules\esbuild\lib\main.js:679:7)
    at Socket.emit (node:events:513:28)
    at addChunk (node:internal/streams/readable:315:12) {
  errors: [Getter/Setter],
  warnings: [Getter/Setter]
}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
ERROR: "build:esbuild" exited with 1.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I solved my issue by cloning the `cdt-gdb-adapter` project to my local and building the `cdt-gdb-adapter` project with the following command:

`$ yarn build && yarn nativebuild`

Then use `yarn link` to link my `cdt-gdb-adapter` project into `cdt-gdb-vscode` project. After that I had chance to solve the missing file issue and successfully build the `cdt-gdb-vscode` project. I just want to mention this in case it is escaped from your attention in development environment, or please guide me if you think I am doing something wrong.

Another issue is related to change in `dist` and `out` folders for output of the Memory Browser. I am still investigating the problem on my side. But it would be good to understand future plans of cdt-gdb-vscode if there is a plan to move all the build output to `out` folder or still merge them under `dist` folder in the future. 

Kind regards.

colin-grant-work:
@asimgunes, thanks for the feedback. I haven't seen the errors you mentioned, but I'll take a look. ESBuild should be able to find that file where it resides and include it in the bundle, but it appears that that didn't work in your case, and I'll see if I can determine the cause.

Re: `out` vs. `dist`, this can be rearranged if it's found that the current setup isn't desirable for some reason - please let me know the results of your investigations.

As you can see, I marked this PR as a proposal. I believe that packaging the plugin will be beneficial, but exactly how it's done is completely open for discussion, of course. If it causes more trouble than it's worth at the moment, I'm completely happy reverting it and discussing the way forward.

colin-grant-work:
Re: `out` vs. `dist`, please note the changes [here](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/96/files#diff-265387a774485ec42c99bc6f655920c93d48b0e43849cf381bfc4ab8e6c64b1f) to the MemoryServer.ts file that ensure that the sources are loaded correctly into the webview.

colin-grant-work:
From your error message, I see that you're on Windows, and I was able to reproduce the error with the `pty.node` file on Windows. It seems that the adapter does no native build at all on Windows since the `pty` code is intended for use only on Linux.

@jonahgraham, how is that currently handled in the publication of the plugin? Are different versions published on different platforms, or is a Linux build shipped to everybody, which is fine in this case, because the code wouldn't even try to be used on Windows?

jonahgraham:
@colin-grant-work I don't publish the extension - someone set it up on open-vsx, but I don't know who did that, nor how it is done. I would assume it is built on Linux though.

The adapter itself builds fine on Windows (the GitHub actions CI checks that), is there something about how it is packaged that the build no longer uses this:

https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/92bb15046fea82256742a69f0b240129a1949a76/install.js#L5-L10

jonahgraham:
> > I've had some more time to confirm that this works downstream, if you're still alright with its merge, @jonahgraham?
> 
> Yes.
> 
> Please (as a separate commit?) do a yarn version to bump version so this change gets published to open vsx (or please ask me and I can do that part)

@colin-grant-work @asimgunes - are we ok with current state to publish to open-vsx? Because we haven't tagged a new version we are still publishing the pre-this PR version to open vsx and that means that users don't have access to recent bug fixes (done in #100)

colin-grant-work:
> @colin-grant-work @asimgunes - are we ok with current state to publish to open-vsx? Because we haven't tagged a new version we are still publishing the pre-this PR version to open vsx and that means that users don't have access to recent bug fixes (done in https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/100)

Ok from our side. Happy to change the TS build to include source maps etc. if that would make treating it as an NPM-ish package easier.

jonahgraham:
IIUC I can publish what we have now (as v0.0.104) and leave the source maps for a future commit?

colin-grant-work:
From our perspective yes. We're also happy without source maps, just offering that as an option if it makes @asimgunes use case easier.

jonahgraham:
Thanks @colin-grant-work.

Done ([tag](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/releases/tag/v0.0.104) and commit).  The new version should end up on openvsx overnight tonight.

PR #98, opened 2023-09-26
Don't include pty.node unless on Linux

colin-grant-work:
The CDT GDB Adapter's [`install.js`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/33b2f35173f67563ebb1d42a6cd4aaf7484cb90c/install.js#L5) only runs the native build if the platform is Linux. That means that in other cases, there will be no `pty.node` file created, and we should not try to include it any bundle generated on a different platform. Alternatively, we could modify the adapter to run the build on other platforms, even if it is not anticipated that the code will be used.

This allows the plugin to be built smoothly on platforms other than Linux.

CC: @asimgunes

### DIFF
--- a/esbuild.js
+++ b/esbuild.js
@@ -1,6 +1,7 @@
 // @ts-check
 const esbuild = require('esbuild');
 const path = require('node:path');
+const os = require('node:os');
 const { sassPlugin } = require('esbuild-sass-plugin');
 
 /** @typedef {import('esbuild').BuildOptions} BuildOptions */
@@ -48,7 +49,7 @@ const configurations = [
             path.join(debugAdapterRoot, 'debugTargetAdapter.js'),
         ],
         outdir: path.join(__dirname, 'dist'),
-        external: ['process'], // Workaround pending https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/288
+        external: os.platform() !== 'linux' ? ['*/pty.node'] : undefined,
         loader: { '.node': 'copy' },
         format: 'cjs',
         platform: 'node',


jonahgraham:
@asimgunes can you test if this resolves the Renesas use case? If so then I think @colin-grant-work can merge it soon. 

@colin-grant-work Asim is on UK timezone, so I wouldn't expect an answer until tomorrow. If he confirms it works, are you ok with me merging this?

colin-grant-work:
> If he confirms it works, are you ok with me merging this?

Yes, by all means.

colin-grant-work:
@asimgunes, sure, I see no problem including the declaration files. I'm a little confused about how you consume them, though. Since this project isn't available as an NPM module it isn't a `node_module`, and it sounds like you fork the project, since you're running the build process defined here, so don't you have access to the TypeScript sources themselves? Or are you consuming the declaration files in a separate project?

jonahgraham:
We have discussed in the past publishing cdt-gdb-vscode as an npm package - would that resolve this issue?

@asimgunes do you have a solution for your current upcoming release? Can you build against/with the v0.0.103 tag?

colin-grant-work:
> @colin-grant-work it is not forked but cloned and have a minor tsconfig.json change before using the project, which is including the declarations. With help of `yarn link` we are adding the project as reference. Adding definitions will ease the steps and that we can directly use the project.

With the `yarn link` step, you're effectively treating it as though it were an NPM dependency, so it does sound like actually publishing it as such would serve your ends. I'm a bit confused why you bother with the compilation step on the plugin, though. I think you could also just use the sources directly with the same effect, for example by placing the plugin sources in a subdirectory of your source tree, since you're getting them as source anyway.

@asimgunes, would it be helpful to set `declarations: true` in the TSConfigs? That way, as part of your building process, if you decide to continue using that, you could just run the `build:tsc` script and then import from the `out` directory, where both declarations and transpiled code would be found?

jonahgraham:
> @jonahgraham, I am not sure about the previous reasons which is preventing cdt-gdb-vscode to be published as an npm package, 

@asimgunes  AFAIR the only reason is no one has done it, if it was needed it probably fell off a priority list. I'll let you and @colin-grant-work decide what is needed here, and if publishing this to npm is the result please let me know and I will start doing that.

colin-grant-work:
> Hi @colin-grant-work, adding the following configuration is solving my problem,
> 
> ```
> "declarationDir": "dist",
> "declaration": true
> ```

The errors you're seeing are because of the `declarationDir: "dist"` configuration, which is outputing to place it doesn't think is an output directory, so it interprets the contents as source rather than output, and it doesn't want to overwrite source. My understanding is this:
 
Formerly, your setup worked because the `main` field of the `package.json` pointed to `out/extension.js`, and `.d.ts` files were also present in out. This means that an import like `import {MemoryServer} from 'cdt-gdb-vscode';` imported from `out/extension.js` and got its types from `out/extension.d.ts`.

Now, `main` poinst to `dist/extension.js`, and there are no `.d.ts` files there, although things like the `MemoryServer` are still exported from there. 

What I'm suggesting:
 - We add  `declaration: true` back into the TSConfigs. 
 - You update your important statements to `import {MemoryServer} from 'cdt-gdb-vscode/out/extension'`
Then you have access to all of the declaration files you need and all of the imports and behavior you're used to. Basically, we treat the `out` directory as the NPM package and the `dist` directory as the plugin bundling target. Locally, rather than running `yarn build` you could just run `yarn build:tsc` to get the TSC compilation output in `out` that you need. The changes in the imports are unfortunately required because the `package.json` `main` field is a keyword for both module resolution & VSCode plugin entry points. Maybe we could even resolve that with the `module` field, but that's supposed to be the export point for ESM modules.

jonahgraham:
Lets continue the remaining issues in new bugs/PRs. I'll merge this now.

PR #107, opened 2023-10-31
Packaging updates

asimgunes:
Hi @jonahgraham, 

I like to discuss this update as a follow-up changes for previous work at pull-request #96.

In this update:

* Removed the "CLI development" option from build scripts, and minify and extract source map at all times.
* Removed the tsc compiler execution at build phase, leave build operation to ESBuild. Thus remove `out` folder extract all the output at `dist` folder.
* Optimised the build scripts and include ESModule format besides to CommonJS format.
* Added types decleration to build output.
* Set the CJS, ESM entry points and types decleration entry points in package.json
* Minor changes at VSIX package files (at .vscodeignore).
* Minor changes at Node package files (at package.json).

Benefits: 
* Included source map files at all times, ease the debug operation.
* Included type declarations made `cdt-gdb-vscode` easy to use as a package depedency.
* Removing tsc at the build phase do not have any side effect (as far as I observed). Besides remove the redundant/unnecessary `out` folder from the build output (Since entry points already changed to `dist` folder.)

I didnot add any script definition for triggering package publishing operation in npm registry, but, after this update, I believe we can publish cdt-gdb-vscode as an npm package and anyone could easily use cdt-gdb-vscode as a dependent package and extend behaviour.

I hope this would create a positive impact for cdt-gdb-vscode.

Kind regards.
Asim

### DIFF
--- a/esbuild.js
+++ b/esbuild.js
@@ -1,21 +1,20 @@
 // @ts-check
 const esbuild = require('esbuild');
 const path = require('node:path');
-const os = require('node:os');
 const { sassPlugin } = require('esbuild-sass-plugin');
 
 /** @typedef {import('esbuild').BuildOptions} BuildOptions */
 
 const args = process.argv.slice(2);
 const CLIWatch = args.includes('--watch');
-const CLIDevelopment = CLIWatch || args.includes('--development');
+const NoMinify = args.includes('--no-minify');
 
 /** @type {BuildOptions} */
 const commonConfig = {
     bundle: true,
     target: ['es2020'],
-    minify: !CLIDevelopment,
-    sourcemap: CLIDevelopment,
+    minify: !NoMinify,
+    sourcemap: true,
 };
 
 const srcDir = path.join(__dirname, 'src');
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
   },
   "homepage": "https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode#readme",
   "main": "./dist/extension",
+  "types": "./dist/extension.d.ts",
   "engines": {
     "vscode": "^1.78.0"
   },
@@ -770,6 +771,12 @@
     "CONTRIBUTING.md",
     "dist/**/*.js",
     "dist/**/*.js.map",
-    "dist/**/*.d.ts"
+    "dist/**/*.css",
+    "dist/**/*.css.map",
+    "dist/**/*.d.ts",
+    "dist/**/*.d.ts.map",
+    "src/**/*.ts",
+    "src/**/*.tsx",
+    "src/**/*.scss"
   ]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -8,6 +8,8 @@
     "lib": ["es2015", "dom"],
     "jsx": "react",
     "strict": true,
+    "declaration": true,
+    "declarationMap": true,
     "noImplicitAny": true,
     "noImplicitReturns": true,
     "noUnusedLocals": true,
@@ -17,5 +19,5 @@
     "preserveWatchOutput": true
   },
   "references": [{ "path": "src/memory/client/tsconfig.json" }],
-  "exclude": ["**/client/*"]
+  "exclude": ["**/client/*", "**/dist/*"]
 }


esbuild.js
```diff
@@ -3,22 +3,22 @@ const esbuild = require('esbuild');
...
-const CLIDevelopment = CLIWatch || args.includes('--development');
 
 /** @type {BuildOptions} */
 const commonConfig = {
     bundle: true,
     target: ['es2020'],
-    minify: !CLIDevelopment,
-    sourcemap: CLIDevelopment,
+    minify: true,
+    sourcemap: true,
```
colin-grant-work, line 18: I would prefer to keep the option not to minify. Source maps help, but it can be useful to inspect the generated JS, as well.

jonahgraham (approved):
This LGTM - but it would be nice if @colin-grant-work can have a look too please.

package.json
```diff
@@ -722,8 +724,6 @@
     }
   },
   "scripts": {
-    "build:tsc": "tsc -b",
-    "watch:tsc": "tsc -b -w",
```
colin-grant-work, line 726: Removing the TSC component of the build appears to remove type checking from the build entirely. I'd prefer to have script option that does perform type checking, though I don't oppose having a build option that does not require type checking, in addition.

However, if we remove the `tsc` components of the build, then the `build` and `watch` scripts can be simplified considerably: no need for the separate `build:esbuild` script or the `pre-watch` and `do-watch` scripts.

.vscodeignore
```diff
@@ -4,7 +4,8 @@
 
 # Include the following:
 !images/
-!dist/
+!dist/**/*.js
+!dist/**/*.css
```
colin-grant-work, line 8: This seems odd. As part of the plugin packaging, this includes both the Common JS entry point and an ESM export, but not the `.d.ts` files that would help users consume those. If the intention is to include in the plugin package only the relevant JS files, then I believe that only `.cjs.js` should be included; if the goal is to include all the files necessary to consume the plugin as an NPM-like dependency, then `.d.ts` files (and maybe the .esm.js file?) should be included, as well.

esbuild.js
```diff
@@ -28,15 +28,24 @@ const debugAdapterRoot = path.join(
...
         format: 'cjs',
         platform: 'node',
         ...commonConfig,
     },
     {
-        entryPoints: [path.join(outDir, 'memory', 'client', 'index.js')],
+        entryPoints: [path.join(sourceRoot, 'extension.ts')],
+        outfile: path.join(__dirname, 'dist', 'extension.esm.js'),
+        external: ['vscode'],
+        plugins: [dtsPlugin()],
```
colin-grant-work, line 42: This plugin's behavior was a little inconsistent for me locally. After pulling the branch and running the build, I got the `.d.ts` files in the `dist` directory as expected. On subsequent builds after doing `git clean -xfd && yarn && yarn build` I didn't get any `.d.ts` files. I'm looking into whether metadata is stored somewhere to try to avoid redundant builds, but any guidance on what could be going wrong there would be helpful.

colin-grant-work: Seems to correspond to [this issue](https://github.com/Floffah/esbuild-plugin-d.ts/issues/8) in the plugin's repo.

colin-grant-work:
This style of export also doesn't play very nicely with the TypeScript language server. In  a dependent, I can write an import for the `MemoryServer` in three ways, all of which the TS server is happy with:

```typescript
import { MemoryServer } from 'cdt-gdb-vscode'; // Will work
```
```typescript
import { MemoryServer } from 'cdt-gdb-vscode/dist/extension'; // Will work
```
```typescript
import { MemoryServer } from 'cdt-gdb-vscode/dist/memory/server/MemoryServer'; // Won't work
```

The problem is that, when the code is bundled, the generated `.d.ts` files don't actually correspond to files present on disk, but the TS server assumes that `.d.ts` files should be in a 1:1 correspondence with transpiled `.js` files, but with this setup, they aren't: it would be more correct to concatenate the content of all of the `.d.ts` files into `extension.d.ts`, since that's the only `.js` file actually available. Perhaps it would be best to run three builds:
 - bundled into `extension.js` for consumption as a plugin
 - unbundled into `.esm.js` files for consumption as ES Modules
 - unbundled into `.cjs.js` file for consumption as Common JS modules.

.vscodeignore
asimgunes: Hi @colin-grant-work, I included `d.ts` files to npm package and consider VSIX package as a runtime environment and exclude the `d.ts` files as well as `js.map` files from the VSIX package. Does this sounds logical to you?

jonahgraham:
@asimgunes please see conversation in #112 also.

asimgunes:
Hi @jonahgraham, @colin-grant-work,

I rebased the request and keep minimum by only adding required type declerations, since @colin-grant-work rollback previous changes.

Could you please review the changes again?

jonahgraham (approved):
It looks fine to me - but I would appreciate @colin-grant-work's review as he understands this area better.

colin-grant-work (approved):
The changes look good to me. 👍

GitMensch:
Nice, this looks good so.. included soon?

PR #116, opened 2023-11-29
Add definitionPath property to debug configuration attributes

FlorentVSTM:
This PR adds the property 'definitionPath' to the attributes of the 'gdbtarget' debug configuration. This property comes from the peripheral-inspector extension.
Adding this property promotes the use of the eclipse-cdt.peripheral-inspector extension and avoids the yellow squiggly line in the launch.json file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -300,6 +300,11 @@
                 },
                 "default": []
               },
+              "definitionPath": {
+                "type": "string",
+                "description": "Setting for 'PERIPHERALS' window provided by 'eclipse-cdt.peripheral-inspector'. It can be a simple file name or based on a CMSIS pack or deviceName. See 'eclipse-cdt.peripheral-inspector' for format",
+                "default": null
+              },
               "imageAndSymbols": {
                 "type": "object",
                 "default": {},
@@ -540,6 +545,11 @@
                   "type": "string"
                 },
                 "default": []
+              },
+              "definitionPath": {
+                "type": "string",
+                "description": "Setting for 'PERIPHERALS' window provided by 'eclipse-cdt.peripheral-inspector'. It can be a simple file name or based on a CMSIS pack or deviceName. See 'eclipse-cdt.peripheral-inspector' for format",
+                "default": null
               },
               "imageAndSymbols": {
                 "type": "object",


colin-grant-work:
@FlorentVSTM, can you say a little bit more about how the two extensions interact? How does the `eclipse-cdt.peripheral-inspector` consume the setting for a `gdb` debug configuration?

FlorentVSTM:
> @FlorentVSTM, can you say a little bit more about how the two extensions interact? How does the `eclipse-cdt.peripheral-inspector` consume the setting for a `gdb` debug configuration?

Hi @colin-grant-work,

yes of course, The peripheral inspector consumes the definitionPath option from the gdb debug configuration thanks to the debug session provided by the tracker "onWillStartSession" (periperalTreeProvider.ts).
Then, in svd-resolver.ts the resolve method checks if the debug configuration of this debug session contains the definitionPath option for use.
Note that users can change the name of the definitionPath option in the extension settings, but by default, it is definitionPath.

colin-grant-work:
The use case sounds reasonable, but I'm not sure the approach is sound, because I'm not sure that it would make sense to document in _this_ plugin all the data that any variant of GDB debugging might use: it should document the schema expected by the CDT-GDB Adapter, whose functionality it exposes to VSCode. On the other hand, I'm not aware of a mechanism in the VSCode API that would allow another plugin to modify the schema of the debug configuration declared here without supplying a debug configuration of its own, which also isn't ideal.

Let me take a look at the code that handles debug schemas and see if we have any options that would allow for a cleaner separation.

FlorentVSTM:
I agree with you that it would not make sense to document in this plugin all the data that any variant of GDB debugging might use. It should primarily document the schema expected by the CDT-GDB Adapter. 
However, if an option is also required by other extensions in the debug configuration and is not documented, then this option will not be suggested via "Ctrl+Space" in launch.json and will be underlined with a yellow squiggly line. So the user must know the exact name of the option and the expected type of the value.
My vision is that to promote and facilitate the debug extensions provided by Eclipse Cloud, it could be interesting to document only those as well (perhaps on a case-by-case basis, if necessary). Of course, they should not be added to the snippets but only documented.

FlorentVSTM:
Hi @colin-grant-work,
I think I found a better way to add this attribute. We can extend the JSON schema defined in cdt-gdb-vscode directly from the vscode-peripheral-inspector extension so that the "definitionPath" attribute is available in the schema only if the vscode-peripheral-inspector extension is installed. 
If you think this leads to cleaner separation, I can propose a pull request in the peripheral-inspector repo and check with Rob.

colin-grant-work:
> We can extend the JSON schema defined in cdt-gdb-vscode directly from the vscode-peripheral-inspector extension so that the "definitionPath" attribute is available in the schema only if the vscode-peripheral-inspector extension is installed.

Perfect - this is exactly what I was hoping for. I apologize that I didn't get around to finding it myself, but I'm glad to know it's possible.

Issue #119, opened 2024-02-19
Publish to the VS Code marketplace


thegecko:
This doesn't seem to have been published to the main VS Code marketplace, any reason why it shouldn't be?

spoenemann:
See also #40

jonahgraham:
No objection - quite the opposite.

As mentioned in the [last cdt cloud call](https://github.com/eclipse-cdt-cloud/cdt-cloud/wiki/CDT-Cloud-Meetings#agenda), the build infra here is quite sorely out of date, this is what I summarized at the call, but I don't know when I will have time to do this:

> This highlighted that the extension (cdt-gdb-vscode) is quite dated in technology and has failed to publish recently. I (Jonah) am considering doing a whole-scale update, which will mean another major version bump to highlight that change.

thegecko:
I'm hoping it should be quite easy to copy the CI from one of our other extensions.

jonahgraham:
> Yes, my intention is to start from existing extensions like the "Inspector" extensions. Unless you have something else in mind. We can then discuss on the PR if this is sufficient or should be further refined/provide more functionality.

Sounds good to me. I have nothing else in mind.

> I might just need some assistance with setting up secrets as I certainly don't have the required rights to do so myself.

I am hoping @thegecko can handle this part, but lmk if you need me for any of it.

thegecko:
We have org wide secrets set up for publishing all extensions, copying another CI workflow should 'just work'.

jreineckearm:
> We have org wide secrets set up for publishing all extensions, copying another CI workflow should 'just work'.

Nice!

thegecko:
Should be closed in #130

jreineckearm:
The good news is that the `CI` Github action started kicking after merging #130 . So, we can test this now properly on branches.
The bad news is that the latest `vsce` version doesn't like node 14 but requires version 20. I'll see how far I get with bumping the node version. If things fall apart with that node update we need to worst case roll back to https://github.com/microsoft/vscode-vsce/releases/tag/v2.15.0 which is the last version with node 14 support.

PR #130, opened 2025-01-10
Add Github actions for build and publish to Open-VSIX and Visual Studio Code Marketplace

jreineckearm:
**Changes**

This PR addresses https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/119 and adds Github Actions for
- Build
- Publish to Open-VSIX
- Publish to Visual Studio Code Marketplace

Github actions have been reused from https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/blob/915701a0b5742e218e893cfa0869a48b21005c8a/.github/workflows/ci.yml
Node version has been changed to 14.x to reflect the currently included node types dev dependency. But this can be easily changed to a newer version.

Other changes in this PR
- Added a clean script.
- Bumped vsce version to latest.
- Amended `.vscodeignore` list. Might need another update to avoid shipping the full node_modules folder including dev dependencies. But should be done separately and with more understanding of production dependencies.

**Note**: The actions have not been tested yet in this repo! The `.vscodeignore` updates have been tested locally.
- ~The build action should run on PR pushes unless I forgot something. So, hopefully gets tested after opening this PR.~ The build action requires updates to this repo's CI which I can't add myself. And presumably will break other PRs until this one here is merged. @thegecko, @jonahgraham I need help here as I don't have required rights to make the changes.
- Publishing itself however cannot be easily tested without making a real release. The actions themselves don't contain URLs or other code with repo specifics. So confidence is high that they work as on the other repos.

**How to publish releases**

In order to make a release, you need to
- Manually update the version in `package.json` as before.
- Then create and push a release version (git) tag to the `main` branch. This should start the release actions to both Open-VSIX and the VS Code Marketplace. Also, a release is created on this Github repo which should include the VSIX file as asset.

**Findings while working on this**

- For other extensions we use a new CDT Cloud icon/image: https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/blob/915701a0b5742e218e893cfa0869a48b21005c8a/media/cdtcloud.png
  - Should we update this for the CDT GDB Adapter extension as well?
- The repository doesn't have a CHANGELOG.md to show the release history on both Open-VSIX and the marketplace. It would be valuable to add that. Happy to make it part of this or a new PR, starting to fill it with life from the next release on.
- As indicated in the discussion in the linked issue, dependencies need a refresh. But we probably want to do this as one or more separate PRs as this certainly requires more testing. And could be done after an initial release to the VS Code marketplace.
- As mentioned further above, we should investigate if we really need to ship the entire `node_modules` folder.


thegecko (approved):
this LGTM

jonahgraham (approved):
+1 from me. 

@jreineckearm, as @thegecko has full rights on here and you are both in the same timezone, can @thegecko merge and work with him on coordinating the various updates? I don't want to be a blocker by being in the loop in unneeded ways. In other words I give my blanket approval for #119 related changes.

If there is concern that breaking changes are being made, bumping the major version to 2.0 would make sense to me so that consumers know to be aware of unexpected changes.

jreineckearm:
@spoenemann , I think this would replace the automated publishing process. The idea is that the tag-push sets things off.
But good point raised, we should ensure that the automated publishing is stopped before making a release to avoid potential conflicts.

Issue #136, opened 2025-01-20
v2.0.0 Release


jreineckearm:
We should make a release ~v1.1.0~ v2.0.0 to publish the current state to both Open VSX and the VS Code marketplace.

Changes since 1.0.4 release:
- CI updates
- Documentation updates
- CDT Cloud logo update

Dependency updates NOT addressed in this release
- Main production dependency to update is React (17.x -> 19.x). It comes with a mild risk of compatibility issues. But it's only used for the Memory Browser which is under discussion to be removed (#110). If it gets removed, the React dependency goes away, too. So, probably not worth the testing effort now.
- Typescript version.
- Eslint version update to 9.x. This is not straight forward due to hard requirement to update configuration files to the new flat configs. Those updates are not trivial (at least not for me) and may take a little more time to get things right.

TODO before release:
* One remaining docs fix.
* Review release notes and update version number in CHANGELOG.md.
* Bump version and add missing properties in package.json (`displayName`). Use descriptions and name in line with [brand guidelines](https://code.visualstudio.com/brand).

[Update]: Targeting now v2.0.0 instead of v1.1.0 release due to displayed extension name changes on marketplace/open vsix registry.

jonahgraham:
> Dependency updates NOT addressed in this release

I think that all dependency updates should be applied to cdt-gdb-adapter too so that the same versions of eslint and typescript are used for both parts of this project.

Somewhat related, if there is some simplifications that can be applied to how the dependency between cdt-gdb-adapter and cdt-gdb-vscode works, that would be great. Now to push out any update to cdt-gdb-adapter (where most of the heavy lifting is done) both the adapter and the extension need work, such as done in https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/120

jreineckearm:
> I think that all dependency updates should be applied to cdt-gdb-adapter too so that the same versions of eslint and typescript are used for both parts of this project.
> 
> Somewhat related, if there is some simplifications that can be applied to how the dependency between cdt-gdb-adapter and cdt-gdb-vscode works, that would be great. Now to push out any update to cdt-gdb-adapter (where most of the heavy lifting is done) both the adapter and the extension need work, such as done in [#120](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/120)


Makes sense. Even more so reason to spend some time and thought on this after this release.

jreineckearm:
[...]

PR #137, opened 2025-01-20
v2.0.0 preparations

jreineckearm:
Addresses #136 (see issue for rational why not updating more dependencies first).
Should be merged after #135.

**_Note that this contains some potentially controversial changes concerning the extension's name._**

Preparations for first release to both VS Code marketplace and Open VSX registry.
- Targets ~v1.1.0~ v2.0.0.
- Updates version in `CHANGELOG.md` and `package.json`.
- Adds some missing/useful fields to the extension manifest. This includes `displayName` which is shown on the marketplace and presumably Open VSX instead of `cdt-gdb-vscode`.
  - Set `displayName` to "CDT GDB Debug Adapter Extension"
  - Changed `description` to "CDT GDB debug adapter extension for Visual Studio Code" to better agree with VS Code [brand guidelines](https://code.visualstudio.com/brand).
  - Updated `README.md`, `DEVELOPMENT.md`, and `CONTRIBUTING.md` titles (in the files) accordingly.

Adding `displayName` and changing description and MD title names (in the files) is the potentially controversial part:
- The extension will show now with a different name in Open VSX. And after installation in VS Code.
- This could be a reason for a major version change, i.e. to v2.0.0. But I believe removal of the Memory Browser would be the next major version update due to removal of a feature. Maybe not an issue at all to have two major version increments in short time (if the removal happens soon).

[Update]: Changed to prepare for v2.0.0, no longer v1.1.0.

@jonahgraham , @thegecko , opinions on this would be much appreciated.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Contributing to the CDT GDB Debug Adapter Extension for Visual Studio Code
 
-Thanks for your interest in this project, part of the [Eclipse C/C++ Development Tools project](https://projects.eclipse.org/projects/tools.cdt)
+Thanks for your interest in this project, part of [Eclipse CDT Cloud](https://eclipse.dev/cdt-cloud/).
 
 The source code can be found in the following repository: https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-# Contributing the CDT GDB Debug Adapter Extension for Visual Studio Code
+# Contributing to the CDT GDB Debug Adapter Extension for Visual Studio Code
 
 Thanks for your interest in this project, part of the [Eclipse C/C++ Development Tools project](https://projects.eclipse.org/projects/tools.cdt)
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-# Contributing to VS Code Debug Extension for GDB
+# Contributing the CDT GDB Debug Adapter Extension for Visual Studio Code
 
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -1,4 +1,4 @@
-# Contributing to VS Code Debug Extension for GDB
+# Contributing to the CDT GDB Debug Adapter Extension for Visual Studio Code
```
jonahgraham, line 1: BTW I noticed that the next line still refers to Eclipse CDT instead of CDT cloud. Probably should match https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/?tab=readme-ov-file#contributing

jonahgraham (approved):
> * The extension will show now with a different name in Open VSX. And after installation in VS Code.

+1 to update the name to this

> * This could be a reason for a major version change, i.e. to v2.0.0. But I believe removal of the Memory Browser would be the next major version update due to removal of a feature. Maybe not an issue at all to have two major version increments in short time (if the removal happens soon).

I think you should jump to 2.0, we can have a 3.0 soon after too. But I don't feel strongly that the changes so far necessitate the change from 1.1 -> 2.0

CONTRIBUTING.md
jreineckearm: Good spot! Will change.

jreineckearm: Fixed

jreineckearm:
Should anything else change on this PR before we can get serious with making a release?

jonahgraham (approved):
I renew my approval.

jreineckearm:
Thanks! @thegecko , would you mind merging this PR and then push a `v2.0.0` tag to `main` to kick off a release?

thegecko:
https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/actions/runs/12927145780

PR #138, opened 2025-01-23
cleanup vscode references

GitMensch:
"Visual Studio Debug Extension" sounds like it is usable with "Visual Studio (IDE)", but that can be easily simplified.

In the other cases I've used "vscode" which is the free software behind the binary-only product "Microsoft Visual Studio Code". Nice side effect - it is more concise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
-# CDT GDB Debug Adapter Extension for Visual Studio Code
+# CDT GDB Debug Adapter Extension for vscode
 
-This is a Visual Studio Code extension that supports debugging using gdb and any other debugger that supports the MI protocol. It is built by the experts that provide the gdb support in the Eclipse C/C++ IDE (CDT).
+This is an extension for vscode that supports debugging using gdb and any other debugger that supports the MI protocol. It is built by the experts that provide the gdb support in the Eclipse C/C++ IDE (CDT).
 
-This extension provides a number of features that integrate into the Visual Studio Code debug environment. This includes launch types, support for the standard debug views as well as a custom viewer for memory browsing.
+This extension provides a number of features that integrate into the vscode debug environment. This includes launch types, support for the standard debug views as well as a custom viewer for memory browsing.
 
 ## Prerequisites
 
@@ -12,7 +12,7 @@ External tools are expected to be present on your system depending on the intend
 
 ## Launch Configurations
 
-The Visual Studio Debug Extension for GDB contributes two debugger types:
+This extension contributes two debugger types:
 * `gdb`: Support for **Local GDB Debug**. Launch or attach to an already running application locally on your host machine using GDB.
 * `gdbtarget`: Support for **Remote GDB Debug**. Launch or attach to an already running remote GDB server using GDB.
 
@@ -105,7 +105,7 @@ Settings related to displaying UART output in the debug console. This object can
 
 ## Memory Browser
 
-The extension comes with a Memory Browser window. However, we have plans to deprecate and remove the window in favour of the Eclipse CDT Cloud [Memory Inspector](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector) extension for Visual Studio Code which is available from the [Visual Studio Code Marketplace](https://marketplace.visualstudio.com/items?itemName=eclipse-cdt.memory-inspector) and the [Open VSX Registry](https://open-vsx.org/extension/eclipse-cdt/memory-inspector).
+The extension comes with a Memory Browser window. However, we have plans to deprecate and remove the window in favour of the Eclipse CDT Cloud [Memory Inspector](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector) extension for vscode which is available from the [Microsoft Marketplace](https://marketplace.visualstudio.com/items?itemName=eclipse-cdt.memory-inspector) and the [Open VSX Registry](https://open-vsx.org/extension/eclipse-cdt/memory-inspector).
 
 Any feedback on these plans is welcomed in the discussion in GitHub issue [#110](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/110).
 


jreineckearm:
@GitMensch , thanks for the PR!

One of the reasons for the now used title is Microsoft's brand guidelines which we, I believe, should honor. Especially if publishing now to their Marketplace. https://code.visualstudio.com/brand#brand-name
Note the "please do not say" part. But I think in some places we can indeed relax this a little. :-)

Let me add some comments on the particular changes themselves.

README.md
```diff
@@ -1,8 +1,8 @@
-# CDT GDB Debug Adapter Extension for Visual Studio Code
+# CDT GDB Debug Adapter Extension for vscode
```
jreineckearm, line 1: I believe this should remain "Visual Studio Code" (first time usage in this document).

README.md
```diff
@@ -1,8 +1,8 @@
...
 
-This is a Visual Studio Code extension that supports debugging using gdb and any other debugger that supports the MI protocol. It is built by the experts that provide the gdb support in the Eclipse C/C++ IDE (CDT).
+This is an extension for vscode that supports debugging using gdb and any other debugger that supports the MI protocol. It is built by the experts that provide the gdb support in the Eclipse C/C++ IDE (CDT).
```
jreineckearm, line 3: Should probably say `VS Code` instead.

README.md
```diff
@@ -1,8 +1,8 @@
...
 
-This extension provides a number of features that integrate into the Visual Studio Code debug environment. This includes launch types, support for the standard debug views as well as a custom viewer for memory browsing.
+This extension provides a number of features that integrate into the vscode debug environment. This includes launch types, support for the standard debug views as well as a custom viewer for memory browsing.
```
jreineckearm, line 5: Same as above: `VS Code`

README.md
```diff
@@ -12,7 +12,7 @@ External tools are expected to be present on your system depending on the intend
 
 ## Launch Configurations
 
-The Visual Studio Debug Extension for GDB contributes two debugger types:
+This extension contributes two debugger types:
```
jreineckearm, line 15: 👍 , ok. Probably overdid it here a little.

README.md
```diff
@@ -105,7 +105,7 @@ Settings related to displaying UART output in the debug console. This object can
 
 ## Memory Browser
 
-The extension comes with a Memory Browser window. However, we have plans to deprecate and remove the window in favour of the Eclipse CDT Cloud [Memory Inspector](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector) extension for Visual Studio Code which is available from the [Visual Studio Code Marketplace](https://marketplace.visualstudio.com/items?itemName=eclipse-cdt.memory-inspector) and the [Open VSX Registry](https://open-vsx.org/extension/eclipse-cdt/memory-inspector).
+The extension comes with a Memory Browser window. However, we have plans to deprecate and remove the window in favour of the Eclipse CDT Cloud [Memory Inspector](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector) extension for vscode which is available from the [Microsoft Marketplace](https://marketplace.visualstudio.com/items?itemName=eclipse-cdt.memory-inspector) and the [Open VSX Registry](https://open-vsx.org/extension/eclipse-cdt/memory-inspector).
```
jreineckearm, line 108: Again, better `VS Code`.
I believe we should at least say `VS Code Marketplace` (with or without `Microsoft`). Microsoft has other marketplaces as well, e.g. https://azuremarketplace.microsoft.com/de-de/

README.md
GitMensch: Possibly... note that by doing so, you use a Microsoft registered trademark (both the long and the short "VS Code" are) for something that is not directly invoked with that binary distribution. And of course **when doing so** the trademark's owner has something to say about _how_ you may use it (all the icons and names stuff in your referenced doc is about that).

From the vscode repo's README:

> [Visual Studio Code](https://code.visualstudio.com/) is a distribution of the Code - OSS [vscode] repository with Microsoft-specific customizations released under a traditional [Microsoft product license](https://code.visualstudio.com/License/).

From their FAQ they link to https://github.com/microsoft/vscode/issues/60#issuecomment-161792005 and make an extra point of https://code.visualstudio.com/docs/supporting/faq#_what-does-built-on-open-source-mean

In my view there are three options:

* go the full MS way using their trademark all over the place (I personally dislike doing that whenever the parts referenced have nothing to do with the trademark
* don't use that at all - because that README will mostly be shown in the editors themselves (extension view), then in the marketplaces - so you are "in" there already, no need to reference it (for GitHub and other places you can additionally talks about "for editors base on vscode including Visual Studio Code" (or however you like it) in the project's description (note, instead of the current "CDT GDB Visual Studio Code Extension" MS seems to want you to say "CDT GDB extension for Visual Studio Code", if you want to reference the trademark directly)
* use the name of the tool that all is based on, which is named "Code - OSS" (until it is re-branded, which is the case for nearly all binary distributions as well as in installations like GitPod) or "vscode"

README.md
GitMensch: ... if you want the trademark name there

README.md
GitMensch: agreed

GitMensch:
Whatever way you choose (thinking about it, dropping the trademarks and names altogether apart from the VS Code Marketplace seems not so bad) - I think there are some general hints on this here.

I _think_ its faster if you do the changes you are ok with, instead of going back and forth in this PR. Feel free to close with the appropriate commit :-)

jreineckearm:
Thanks for the feedback, @GitMensch !
Let me have another thought on this. I start getting some of your points. Also in the light of these extensions targeting the Theia IDE, too. So, something more neutral wherever it is suitable makes sense.
I'll make adjustments to this PR as soon as I can. Would then be keen to hear your thoughts with whatever I come up with.

GitMensch:
closed in favor of #142

PR #139, opened 2025-01-23
Update to adapter v1.0.4

jonahgraham:
Includes these changes from the adapter:
- Add supportsEvaluateForHovers
   - https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/347
- Disassembly address handling improvement
   - https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/348

PR #140, opened 2025-01-23
Prepare to release v2.0.1

jonahgraham:
- Main change is #139

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cdt-gdb-vscode",
-  "version": "2.0.0",
+  "version": "2.0.1",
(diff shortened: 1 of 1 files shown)


jonahgraham:
@thegecko @jreineckearm  I think I have done the steps properly. Please advise if you think I did something wrong. I tested the updated extension from the vscode marketplace and it still seems to work :-)

jreineckearm:
@jonahgraham , apologies for the late response. The one thing missing is the update to the CHANGELOG.md. But we can do this for the next update. I am sure more will come in the foreseeable future. :-)

PR #142, opened 2025-01-24
Remove avoidable VS Code references

jreineckearm:
Removed avoidable VS Code references.
See discussion on original PR #138 .

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-# Contributing to the CDT GDB Debug Adapter Extension for Visual Studio Code
+# Contributing to the CDT GDB Debug Adapter Extension
 
 Thanks for your interest in this project, part of [Eclipse CDT Cloud](https://eclipse.dev/cdt-cloud/).
 
@@ -6,7 +6,7 @@ The source code can be found in the following repository: https://github.com/ecl
 
 ## Project description
 
-This is a Visual Studio Code extension that supports
+This is an extension that supports
 debugging using gdb and any other debugger that supports
 the MI protocol. It is built by the experts that provide
 the gdb support in the Eclipse C/C++ IDE (CDT).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,10 +6,8 @@ The source code can be found in the following repository: https://github.com/ecl
 
 ## Project description
 
-This is an extension that supports
-debugging using gdb and any other debugger that supports
-the MI protocol. It is built by the experts that provide
-the gdb support in the Eclipse C/C++ IDE (CDT).
+This extension supports debugging using GDB and the MI protocol. It is built
+by the experts that provide the GDB support in the Eclipse C/C++ IDE (CDT).
 
 ## Build Instructions
 
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "cdt-gdb-vscode",
   "version": "2.9.0",
   "displayName": "CDT GDB Debug Adapter Extension",
-  "description": "Extension from Eclipse CDT Cloud that integrates local and remote debugging with GDB",
+  "description": "Extension from Eclipse CDT that integrates local and remote debugging with GDB",
   "publisher": "eclipse-cdt",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "cdt-gdb-vscode",
   "version": "2.9.0",
   "displayName": "CDT GDB Debug Adapter Extension",
-  "description": "CDT GDB debug adapter extension for Visual Studio Code",
+  "description": "Extension from Eclipse CDT Cloud that integrates local and remote debugging with GDB",
   "publisher": "eclipse-cdt",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "cdt-gdb-vscode",
   "version": "2.9.0",
   "displayName": "CDT GDB Debug Adapter Extension",
-  "description": "Extension from Eclipse CDT that integrates local and remote debugging with GDB",
+  "description": "Extension from Eclipse CDT that integrates local and remote debugging with GDB in VS Code and compatible IDEs",
   "publisher": "eclipse-cdt",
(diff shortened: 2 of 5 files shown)


jreineckearm:
@GitMensch , please see this PR to follow up on #138. Created a new one to avoid forking from your fork.
@jonahgraham , please also review if you are happy with the changed extension description.

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "cdt-gdb-vscode",
   "version": "2.0.1",
   "displayName": "CDT GDB Debug Adapter Extension",
-  "description": "CDT GDB debug adapter extension for Visual Studio Code",
+  "description": "Extension from Eclipse CDT Cloud that integrates local and remote debugging with GDB",
```
GitMensch, line 5: "Cloud" seems a bit strange here and I would suggest to just drop that single word

GitMensch (approved):
LGTM in general

package.json
jreineckearm: Removed

GitMensch:
@jreineckearm What is the status of this PR?

jreineckearm:
Thanks for the reminder, @GitMensch . Slipped off the radar.... 😕 
Will rebase and review if new references slipped in.

cwalther:
I agree that there are some references that are unnecessary and can be removed, but I find it rather confusing that the README now only calls this an “extension” but never specifies _what_ it extends. One can only guess it from the project name. I think there should be at least one mention of Visual Studio Code somewhere near the beginning, e.g.
> This is an extension that supports debugging using GDB and the MI protocol in Visual Studio Code and compatible IDEs.

or
> This is an extension for Visual Studio Code and compatible IDEs that supports debugging using GDB and the MI protocol.

That seems consistent with Microsoft’s brand guidelines.

Other than that, I have no strong opinions.

GitMensch:
I agree with @cwalther (I _think_ my original PR used "vscode" in at least one place).

I'd have to check, but the DAP part may even be usable without vscode - for example in emacs/vim? Then "Debug Adapter Protocol" may be used in some place as well.

jreineckearm:
I like the wording "Visual Studio Code and compatible IDEs". I'll add that to the first paragraph of the readme tomorrow.
I'll also extend the changed description by "....with GDB into VS Code and compatible IDEs". That should mitigate your original concerns of mistaking this for a Visual Studio extension (not a VS Code extension).

Yes, the original PR used "vscode". But TBF: I'd stay with "VS Code" and similar. At the end of the day we are shipping a VSIX (Visual Studio Extension). So, I am not convinced there is a lot of benefit in introducing subtle differences like saying "vscode". It may rather confuse users who are not that familiar with all these details.

CONTRIBUTING.md
```diff
@@ -1,15 +1,15 @@
...
 The source code can be found in the following repository: https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode
 
 ## Project description
 
-This is a Visual Studio Code extension that supports
-debugging using gdb and any other debugger that supports
-the MI protocol. It is built by the experts that provide
-the gdb support in the Eclipse C/C++ IDE (CDT).
+This is extension that debugging using gdb and any other
+debugger that supports the MI protocol. It is built by
```
cwalther: This is gibberish.

cwalther (changes requested):
Not to mention that “vscode” is against the guidelines (“Note proper capitalization and spacing”).

---

You new commit 008397201dbae43f27916c24f14ff5b7e1c87f68 turned out empty, was that intentional? Seems like a regression (6c3754016b78a990cd205555ed80a88b1dbc5333 was good).

CONTRIBUTING.md
jreineckearm: Duh.....thanks!
Was messing around with squashing commits and wanted to proof-read in the GH frontend, then got interrupted by other queries. Let me fix that.

jreineckearm: Nope, it wasn't the squashing....just realized it was already gibberish when it come out of my brains .... anyway needs a fix. 🙂

jreineckearm:
Text should be fixed. I am planning to squash on merge, hence won't bother with fixing the commit history now.

cwalther:
:+1: on that. I am still missing the change from 6c3754016b78a990cd205555ed80a88b1dbc5333 that I think was good, regardless of commit history. (Or an explanation that you removed it intentionally.)

GitMensch:
> Not to mention that “vscode” is against the guidelines (“Note proper capitalization and spacing”).

The point is exactly that "VS Code" = "Visual Studio Code" ==> the Microsoft _product_ (= software + customizations, partially closed source, often not under a free license), while "vscode" is the MIT licensed _software_ (where people can build free versions from and which the community sends PRs against as well). Extensions are written to the software, not the specific product.

There is also software - like Theia, which are not based on vscode, but provide a _vscode compatible_ extension interface.

--> Visual Studio Code (which possibly should be best referred to with including Microsoft as it is their product under their license).

/rant off

jreineckearm:
@cwalther , must have screwed up that commit while squashing things initially. Now squashed all changes since yesterday into one [commit](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/142/commits/9ecc3e2b49868b2b9b8a0001a332a485a9707991) to move on.

@GitMensch , I get where you are coming from. But even Theia, which is probably the closest to this project, uses Microsoft terminology, not `vscode`. See
- https://theia-ide.org/
- https://theia-ide.org/docs/authoring_vscode_extensions/

Will follow that example to keep consistency.

cwalther:
@jreineckearm, my last comment still stands – I think you missed one commit in your squashing. Please
- either reinstate the change from 6c3754016b78a990cd205555ed80a88b1dbc5333 (merge 9ecc3e2b49868b2b9b8a0001a332a485a9707991 and 6c3754016b78a990cd205555ed80a88b1dbc5333 or rebase one onto the other or equivalent)
- or state that you removed that change deliberately.

jreineckearm:
Change reinstated as https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/142/commits/94ae3856c88a7ab0a3fc334921dd7e3e73885cfd

cwalther (approved):

Issue #144, opened 2025-03-11
Error with the openGdbConsole option


DongNguyenHuy:
Hi, I'm debugging my application on RISC-V QEMU with the cdt-gdb-vscode extension. I want to interact with gdb, so I enabled the openGdbConsole option and got the error:
![Image](https://github.com/user-attachments/assets/cabc5197-141c-4147-bf10-dda77805e4d5)

Here is the configuration:
```
{
            "name": "CDT GDB Debug",
            "type": "gdbtarget",
            "request": "launch",
            "gdb": "riscv64-unknown-elf-gdb",
            "program": "${workspaceFolder}/riscv-probe/build/bin/rv64imac/qemu-sifive_e/alloca",
            "logFile": "${workspaceFolder}/.vscode/gdb.log",
            "target": {
                "type": "remote",
                "host": "localhost",
                "port": "50000",
                "server": "qemu-system-riscv64",
                "serverParameters": [
                    "-nographic",
                    "-machine", "sifive_e",
                    "-bios", "none",
                    "-kernel", "${workspaceFolder}/riscv-probe/build/bin/rv64imac/qemu-sifive_e/alloca",
                    "-gdb", "tcp::50000",
                    "-S",
                    // "-serial", "tcp:localhost:5555,server,nowait"
                ],
                "automaticallyKillServer": true,
                "serverStartupDelay": 500,
                // "uart":{
                //     "socketPort": "5555"
                // }
            },
            "initCommands": [
                "b main"
            ],
            // "verbose": true,
            "openGdbConsole": true,
        },
```

Is there anything that I am missing? How can I interact with the gdb?

Thank you!

jonahgraham:
@jreineckearm I can reproduce this, it looks like something is wrong with the packaging of the extension on vscode marketplace. The native pty.node code is missing. 

If I launch in debugServer mode (i.e. set       `"debugServer": 4711` in launch.json) and launch cdt-gdb-adapter from the source directory (using https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/92bb15046fea82256742a69f0b240129a1949a76/.vscode/launch.json#L4-L12) then it works.

---

@DongNguyenHuy as a workaround until we can resolve this, you can run commands in the Debug Console by prefixing them with `>` - that doesn't work for multi-line commands, but allows interacting with GDB directly.

[...]

Issue #146, opened 2025-03-28
Dev container support


flit:
The following error is reported via an alert window when starting a remote debug session in a Docker container.

Given the error, it probably doesn't matter, but I was attempting to connect to a gdbserver running in an emulated machine all within a Docker container. This setup has been successfully tested with the Native Debug extension.

```
Cannot find module '../../build/Release/pty.node'
Require stack:
- /home/username/.vscode-server/extensions/eclipse-cdt.cdt-gdb-vscode-2.0.1/node_modules/cdt-gdb-adapter/dist/native/pty.js
- /home/usernameusername/.vscode-server/extensions/eclipse-cdt.cdt-gdb-vscode-2.0.1/node_modules/cdt-gdb-adapter/dist/desktop/processManagers/GDBPTYProcessManager.js
- /home/username/.vscode-server/extensions/eclipse-cdt.cdt-gdb-vscode-2.0.1/node_modules/cdt-gdb-adapter/dist/desktop/factories/GDBBackendFactory.js
- /home/username/.vscode-server/extensions/eclipse-cdt.cdt-gdb-vscode-2.0.1/node_modules/cdt-gdb-adapter/dist/desktop/GDBDebugSession.js
- /home/username/.vscode-server/extensions/eclipse-cdt.cdt-gdb-vscode-2.0.1/node_modules/cdt-gdb-adapter/dist/desktop/GDBTargetDebugSession.js
- /home/username/.vscode-server/extensions/eclipse-cdt.cdt-gdb-vscode-2.0.1/node_modules/cdt-gdb-adapter/dist/debugTargetAdapter.js
- /home/username/.vscode-server/bin/6609ac3d66f4eade5cf376d1cb76f13985724bcb/[eval]
```

Debug configuration is as follows:

```json
{
    "name": "debug",
    "type": "gdbtarget",
    "gdb": "/tools/arm-gnu-toolchain-14.2.rel1-aarch64-aarch64-none-linux-gnu/bin/aarch64-none-linux-gnu-gdb",
    "request": "attach",
    "program": "<path-to-executable>",
    "target": {
        "port": "2345",
    },
    "cwd": "${workspaceRoot}",
    "openGdbConsole": true,
}
```

### cdt-gdb-vscode version

2.0.1

### VSCode version

Version: 1.98.0 (Universal)
Commit: 6609ac3d66f4eade5cf376d1cb76f13985724bcb
[...]

PR #148, opened 2025-04-30
#144: CI - do not ignore scripts for yarn install

jreineckearm:
Addresses #144  - fix to be confirmed

Remove ` --ignore-scripts` during `yarn install` in CI.
Required to build dependencies, in particular `nativebuild` script of `cdt-gdb-adapter`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
           node-version: 20.x
       - name: Build
         run: |
-          yarn install --ignore-scripts
+          yarn install
           yarn build
           yarn vsce:package
       - uses: actions/upload-artifact@v4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## 2.0.3
+- Fixes [#144](Error with the openGdbConsole option)
+
 ## 2.0.2
 - Update to cdt-gdb-adapter v1.0.6
   - [Hardware/Software Breakpoint Modes](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/350)
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cdt-gdb-vscode",
-  "version": "2.0.2",
+  "version": "2.0.3",
   "displayName": "CDT GDB Debug Adapter Extension",
   "description": "CDT GDB debug adapter extension for Visual Studio Code",
   "publisher": "eclipse-cdt",


jreineckearm:
Missing folder shows up with
https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/actions/runs/14758616632/artifacts/3038964641

Version and changelog gets bumped with this PR, so patch release could be made right away. I haven't had the chance to fully test the fix though.

jonahgraham:
My question better belongs here:

How do I test the vsix as it is installed from marketplace?

I can easily test the change on Linux, not so much on mac or Windows.

jreineckearm:
I just tested the latest version from marketplace on Windows and I get the message
`cdt-gdb-adapter: openGdbConsole is not supported on this platform`

So, testing on Linux is currently sufficient. It seems it's the only platform that supports the feature. See also
* https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/blob/bc9631b7973ac02c27053cec512b2c2b48d6c24e/package.json#L129
* https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/41e5a219797ace77afe1b5f9150bf156c89d0717/src/gdb/GDBDebugSessionBase.ts#L219

It also explains why the extension always targeted the "Universal" platform in OpenVSX.

Going forward, we may want to look into support on Windows/macOS for the console as well. Needs to be figured out then if keeping a universal extension build that includes native binaries for all platforms is easy to create/maintain. Or if we need to switch to host specific extension builds and uploads.

I don't see it as a must though to fix the regression on Linux. Could you please give it a try on Linux? I am on Windows and don't have a Linux machine at hand.

Please also note that it is a public holiday in Germany tomorrow. And I am only back at work on Monday. I might check in over the long weekend but can't guarantee.

jreineckearm:
Just came across #146 which shows the same problem. But presumably on macOS? Might need another test there. Can ask for help on Monday on our end.

jonahgraham (approved):
Assuming I am reproducing the build reliably on my local machine this is working and good to go. Please ping me when it is on marketplace so I can verify that when installed from there (i.e. via CI) that it works as expected.

jreineckearm:
Oh I see. Sorry, I misunderstood your question.
* You can download the VSIX file built with CI here. That CI artifact runs the same build steps as the final release build.
  * Most recent GH action for this PR (see Assets at the bottom): https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/actions/runs/14758616632
  * or direct link to zipped VSIX: https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/actions/runs/14758616632/artifacts/3038964641
* Unzip the file
* Install the VSIX file through the Extension Manager in VS Code or Theia through `More Actions`:
image

After you are done, you should uninstall through the extension manager and may have to manually remove the extension installation folder from VS Code's extension cache. It normally doesn't overwrite it if you re-install an extension with the same version.

Additionally, you could unzip the VSIX from the action and the latest VSIX from marketplace, and compare their contents. VSIX files are effectively zip archives with a different file ending.
The fixed build contains an additional folder `./node_modules/cdt-gdb-adapter/build/Release` which isn't there for extension versions 2.0.0 - 2.0.2.

jreineckearm:
@jonahgraham , v2.0.3 is released to marketplace and OpenVSX. Please let me know if it still works for you. Then I inform users from related issues.

jonahgraham:
@jreineckearm v2.0.3 installed and worked from VSCode marketplace.

jreineckearm:
Thanks, @jonahgraham !

PR #153, opened 2025-05-14
Adding a custom event to trigger re-rendering of breakpoints in source code

omarArm:
Hello,

I added a custom event for triggering a re-render of breakpoints in source code, because just sending a breakpoint event from the adapter to the extension was not enough

Please see
https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/368


jreineckearm:
Turns out this was an issue in the adapter side of the feature. Fixed with latest commits.

jreineckearm:
Merging, one last test when pulling adapter v1.0.9 before next release.

PR #174, opened 2025-07-24
Allow empty program arg

thorstendb-ARM:
See https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/403


jreineckearm:
Addresses https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/367

jreineckearm (approved):
Changes LGTM!
Covers all combinations but `launch` with type `gdb` which genuinely doesn't make sense without a `program`. Good to merge now, but requires cdt-gdb-adapter 1.1.1 to be released to fully work.

PR #186, opened 2025-10-13
Remove statement about support for "any" MI debugger

### LINKED ISSUE #143: Question: MI protocol compatibility statement in README
I've got a question regarding the statement in the repo's README.MD:

We begin with the statement:
"This is a Visual Studio Code extension that supports debugging using gdb and any other debugger that supports the MI protocol."

Does the "and any other debugger that supports the MI protocol" part still hold true?

The `cdt-gdb-adapter` internals make use of GDB specific command interpreter commands. Not only the MI protocol. And in one place we even bail after a version check if it doesn't contain `gdb`: https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/db80d298ee920d84917576a4331e0065017eecc2/src/util/parseGdbVersionOutput.ts#L21

I believe we need to review that statement unless I miss something.

jreineckearm:
Fixes #143 

- Removes statement that any debugger with MI protocol is supported (we have too many GDB specifics).
- Use capital `GDB` instead of `gdb`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # CDT GDB Debug Adapter Extension for Visual Studio Code
 
-This is a Visual Studio Code extension that supports debugging using gdb and any other debugger that supports the MI protocol. It is built by the experts that provide the gdb support in the Eclipse C/C++ IDE (CDT).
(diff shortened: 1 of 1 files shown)


omarArm (approved):
LGTM

PR #191, opened 2025-11-13
Enable breakpoints in assembler files (.asm, .s, .S)

jreineckearm:
Defines a pseudo-language 'cdt-gdb-asm' which defines expected file endings for assembler files, and adds it to breakpoint support.


omarArm (approved):
LGTM

PR #217, opened 2026-04-23
adding global radix change to context menu in different views

omarArm:


src/switchRadix.ts
```diff
@@ -12,26 +12,65 @@ import * as vscode from 'vscode';
...
+
+    private setCurrentRadixContext(session: vscode.DebugSession | undefined) {
+        if (!session) {
+            return;
+        }
+        // Check if the session is already in the map
+        const existingSessionRadix = this._sessionsMap.get(session.id);
+        if (existingSessionRadix) {
+            vscode.commands.executeCommand('setContext', 'cdt.debug.outputRadix', existingSessionRadix);
+            this.handleSetOutputRadix(existingSessionRadix);
```
cwalther, line 41: Why is this line necessary? In what circumstances is the session’s radix not already set to what we last stored for it? Maybe when the user has manually set it using a debug console command, but in that case I would rather respect the user’s setting than override it.

cwalther (commented):
Looks useful and appears to work (at least with a single session, I didn’t try multiple).

However, the commands also appear in unrelated debug sessions (e.g. debugging JavaScript), is that intended? And when executed there fail with a cryptic error (`Failed to set output radix to hexadecimal: CodeExpectedError: Uncaught SyntaxError SyntaxError: Unexpected token '>'`).

Also, when such an unrelated debug session ends (even when I hadn’t used these commands during it), I get an error notification `Failed to set output radix to decimal: Canceled: Canceled`.

jreineckearm:
Thanks for the review, @cwalther !

We had tested multi-core use cases: the states are correctly stored, and contexts correctly updated when switching between sessions.

Well spotted, that we missed to limit this to `gdb` and `gdbtarget` types. Needs also fixing for other radix-related entries in the contributions.

src/switchRadix.ts
omarArm: the problem is that the "set radix to <radix>" depends entirely on the "known" radix set by the system. If the radix changed via the debug console for debug session A and then the user went from session A to session B and back. The user would then find the commands being shown in the various context menus not aligned with that of the actual radix being set.

As a contingency, I thought of using the `show` command to get the current radix, but I figured that would be expensive for the system performance wise

cwalther: OK, valid point. But why would `show` be more expensive than `set`? That would actually seem to be the nicer solution to me.

Sending GDB commands on something as unrelated to GDB as a context switch in the GUI feels somehow inappropriate to me either way, but I would accept it in return for a better user experience. Following what the user sets manually is a better user experience (although, why only on session switch then, what about users who only have one session?). Overwriting what the user sets manually is a worse user experience. I think I would find that infuriating and the greater evil than seeing the wrong commands in the menus.

This is not a showstopper, feel free to go ahead in whatever way you choose.

cwalther (approved):
I would like to eventually see the hardcoded `(debugType == gdb || debugType == gdbtarget)` replaced by something extensible by subclassers (such as me) with their own debug types, but I can contribute that myself.

src/switchRadix.ts
cwalther: Come to think of it, in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/523 we were using an event to react to changed radices. I would say that _that_ would be the proper way of following what the user sets manually, rather than sending extra `show` or `set` commands. (The GDB event is received in the adapter, so it would have to be passed on to the extenson using a custom DAP event.)

omarArm:
The problem with the `show` command, is that we can't retrieve the result without having to read log outputs from the adapter. Which is, from previous experience, is a pain and not ideal.

I think the way to go is that the adapter should trigger a custom event along with the invalidated event once the radix is changed somehow. Either by an extension command, or via the debug console. That of course would need some work in the adapter that I would be happy to do once this PR is merged

cwalther:
Sounds good to me.

omarArm:
https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/219

I created the aforementioned issue for showcasing next steps

jreineckearm (approved):
Thanks, changes look good to me! Going to merge.

Was thinking about the hard-coded adapter types: guess simplest way would be to create a context variable like `cdt.debug.isSupportedAdapter` which gets updated on each active session change. Then at least the `when` conditions wouldn't need to be touched for adapters derived from `gdb`/`gdbtarget`. But not as part of the radix work.

cwalther:
Thanks for the idea, let’s continue that discussion in #220.

PR #221, opened 2026-05-08
Handle custom radix event

### LINKED ISSUE #219: Handling discrepancies between debug console and context menus when it comes to global radix change
As per the discussion in https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/217 . It has been agreed that the extension needs a custom event to handle users changing radices from the debug console without breaking the current system provided by the GUI for changing radices.

This should be solved in 2 steps

1. Creating a custom event in the cdt-gdb-adapter to be sent, along with an invalidated event, whenever the session's radix changes.
2. This custom event should be handled by the extension and updates the context accordingly.

omarArm:
should fix https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/219

Also, goes hand in hand with https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/534/ in the cdt-gdb-adapter

### DIFF
--- a/src/switchRadix.ts
+++ b/src/switchRadix.ts
@@ -33,7 +33,7 @@ export class SwitchRadix {
                     return {
                         onDidSendMessage: (message: DebugProtocol.Event) => {
                             if (message.event === 'OutputRadixUpdated') {
-                            const radix = message.body.radix === 16 ? 'hexadecimal' : 'decimal';
+                            const radix = message.body.radix === '16' ? 'hexadecimal' : 'decimal';
                             vscode.commands.executeCommand('setContext', 'cdt.debug.outputRadix', radix);
                         }
                     }
--- a/src/switchRadix.ts
+++ b/src/switchRadix.ts
@@ -24,17 +24,22 @@ export class SwitchRadix {
         this.registerDebugAdapterTracker();
     }
 
-    private registerDebugAdapterTracker() {vscode.debug.registerDebugAdapterTrackerFactory('*', {
-            createDebugAdapterTracker(session) {
-                return {
-                    onDidSendMessage: (message) => {
-                        if (message.type === 'event' && message.event === 'OutputRadixUpdated') {
+    private registerDebugAdapterTracker() {
+        const debugTypes = ['gdb', 'gdbtarget'];
+        // Register a debug adapter tracker for each debug type
+        debugTypes.forEach(debugType => {
+            vscode.debug.registerDebugAdapterTrackerFactory(debugType, {
+                createDebugAdapterTracker() {
+                    return {
+                        onDidSendMessage: (message: DebugProtocol.Event) => {
+                            if (message.event === 'OutputRadixUpdated') {
                             const radix = message.body.radix === 16 ? 'hexadecimal' : 'decimal';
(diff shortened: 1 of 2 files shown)


src/switchRadix.ts
```diff
@@ -21,6 +21,26 @@ export class SwitchRadix {
...
+    private registerDebugAdapterTracker() {
+        const debugTypes = ['gdb', 'gdbtarget'];
+        // Register a debug adapter tracker for each debug type
+        debugTypes.forEach(debugType => {
+            vscode.debug.registerDebugAdapterTrackerFactory(debugType, {
+                createDebugAdapterTracker() {
+                    return {
+                        onDidSendMessage: (message: DebugProtocol.Event) => {
+                            if (message.event === 'OutputRadixUpdated') {
+                            const radix = message.body.radix === 16 ? 'hexadecimal' : 'decimal';
```
cwalther: Is `body.radix` a number or a string now? This seems inconsistent with https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/534. Does this work?

Also, this logic doesn’t seem to result in the optimal user experience when `body.radix` is something other than 10 or 16. In that case I would expect both commands to be available in the menus.

cwalther (commented):
Is there a reason why you are using a debugAdapterTracker instead of [onDidReceiveDebugSessionCustomEvent](https://code.visualstudio.com/api/references/vscode-api#debug.onDidReceiveDebugSessionCustomEvent), which would seem to me to be the more standard way (unless I’m confused)?

src/switchRadix.ts
omarArm: I set it to a string now, that was the plan all along.

I haven't thought about the idea of making it a multi-state system since only two commands are available. I guess now if a third radix is set, users will have to switch to decimal and then from there choose whichever they want. I don't think it's the best approach from a usability standpoint, but then I will have to change a few things in the package.json.

Let me test a few things before requesting your review again

omarArm:
@cwalther No specific reason for choosing a debug tracker instead of the event, I just didn't catch the event in the documentation and I went with the approach I usually would use for such an issue. Do you think there might be any disadvantages to that?

cwalther:
I am not aware of any specific disadvantage, but there might be a performance impact as every DAP message needs to go through the DebugAdapterTracker, whereas onDidReceiveDebugSessionCustomEvent hopefully only affects actual custom events (although for all sessions). Also, to me using a DebugAdapterTracker for such a purpose feels like a last-resort low-level escape hatch that one only uses when there is no other more targeted, higher-level way of achieving a goal, but opinions may vary on that.

omarArm:
I guess now it should work for all corner cases. I added a third option called `others`, so that if the radix is set to anything other than hex or decimal, both commands will be available.

Also I switched to using onDidReceiveDebugSessionCustomEvent, I read a bit in the documentation and it seems to be the better solution indeed. Thanks for pointing me to it.

Also, I would like to note that @jreineckearm is away for the upcoming few days, so I would appreciate if you can merge after your approval, as I can't merge myself since I am not a committer

cwalther:
Looks good to me and appears to work, thanks!

PR #223, opened 2026-05-11
Check formatting in CI

cwalther:
It looks like nobody has run `yarn format` on this project for a long time. (I only did because I noticed some inconsistent spacing in a previous PR.) If we still find automatic code formatting useful and provide this command, I propose that we also check it in pull request CI, otherwise the formatting diverges over time.

I have no strong opinion myself.

I took the liberty of adding `"trailingComma": "es5"` because
- we do the same in cdt-gdb-adapter,
- it makes a shorter diff to the existing code style,
- trailing commas in argument lists look weird to me (they are useful in array literals, but argument lists are a different thing).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,10 @@ jobs:
           yarn install
           yarn build
           yarn vsce:package
+      - name: Verify code formatting is valid
+        run: |
+          yarn format-check
+          yarn lint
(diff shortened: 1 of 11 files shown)


cwalther:
(low-prio)

jreineckearm (approved):
All makes perfect sense, thanks for some housekeeping!

Merging while you are away to make sure the changes become active before new PRs come in.

PR #224, opened 2026-05-21
Add source code highlighting

omarArm:
- This PR totally depends on https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/537
- Whenever user switches to a different source file, the system retrieves lines in which users can install breakpoints
- These lines are then highlighted

Note: We can agree on what colour to choose for highlighting, that wouldn't be a major change

image

### DIFF
--- a/src/SourceFileHighlighting.ts
+++ b/src/SourceFileHighlighting.ts
@@ -18,7 +18,7 @@ export class SourceFileHighlighting {
         vscode.window.createTextEditorDecorationType({
             borderWidth: '0 0 0 2px',
             borderStyle: 'solid',
-            borderColor: 'rgba(102, 145, 255, 0.85)',
+            borderColor: new vscode.ThemeColor('editorLineNumber.foreground'),
             isWholeLine: true,
         });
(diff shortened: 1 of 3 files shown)


src/SourceFileHighlighting.ts
```diff
@@ -0,0 +1,121 @@
+/**
+ * Copyright 2026 Arm Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
jreineckearm: Please change license to EPL 2.0 as for the rest of this repo.

omarArm: done

jreineckearm (approved):
Thanks, looks good to add for testing it with users!

Future enhancements:
- add an extension setting to enable/disable the highlighting
- have another look if we can have a wider.

jreineckearm (approved):
Note: I am pushing for a release before I go on holiday ago. If anyone would like to see adjustments, please let @omarArm know.

PR #228, opened 2026-05-27
v2.8.0 release preparation

### LINKED ISSUE #218: Race condition when using breakpoint modes to set HW/SW breaks
Follow-up to PR #211 :

Above PR added context menu entries to explicitly set HW/SW breakpoints opposed to setting a "normal" breakpoint and changing the mode afterwards.

The implementation had to apply some workarounds due to limitations in the VS Code API and its DAP protocol support.
This introduced a race condition between setting/removing breakpoints when the added commands apply the required sequence of setting/modifying a breakpoint programmatically. This caused the following two side effects, of which the removal of all breakpoints is probably more problematic than just seeing unexpected log messages.

Reproducers:
1. Set a breakpoint through the column/margin, 2.) use one of the two new commands => I always see the log output No breakpoint number <n>. Is there anything we can do to suppress this? It's not super-blocking and only happens when explicitly using one of the new commands. But made me look twice to ensure this isn't a result of something going wrong. I expect similar confusion at users' end.
1. Set a breakpoint through the column/margin, 2.) use "set software breakpoint", 3.) call a > info b from the debug console => all breakpoints are gone in GDB (still visible in GUI). I am afraid this needs another look. This doesn't happen when you use "set hardware breakpoint" first.

omarArm:
- Implements [cdt-gdb-adapter `#201`](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/201): Add a command that allow users to change GDB session radix
- Implements [cdt-gdb-adapter `#206`](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/206): Add "Set Output Radix" commands to context menus
- Implements [cdt-gdb-adapter `#210`](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/210): Support value formatting for single expressions
- Implements [cdt-gdb-adapter `#224`](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/224): Add source code highlighting
- Fixes [cdt-gdb-adapter `#218`](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/218): Race condition when using breakpoint modes to set HW/SW breaks
- Update to cdt-gdb-adapter [v1.9.0](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/releases/tag/v1.9.0)
    - Implements [`#527`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/527): Support setting expression values

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,14 @@
 # Change Log
 
-## Unreleased
+## 2.8.0
 
 - Implements [cdt-gdb-adapter `#201`](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/201): Add a command that allow users to change GDB session radix
+- Implements [cdt-gdb-adapter `#206`](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/206): Add "Set Output Radix" commands to context menus
+- Implements [cdt-gdb-adapter `#210`](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/210): Support value formatting for single expressions
+- Implements [cdt-gdb-adapter `#224`](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/224): Add source code highlighting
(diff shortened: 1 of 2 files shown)


jreineckearm (approved):
Thanks, LGTM!
Did a quick smoke test.

PR #237, opened 2026-06-22
Source code highlighting cleanup

### LINKED ISSUE #232: SourceFileHighlighting sends breakpointLocations requests to non-GDB debug sessions, breaking ms-python.debugpy (regression in v2.8.0)
Hi @omarArm and @jreineckearm ,

The **SourceFileHighlighting** class (introduced in v2.8.0) sends **breakpointLocations** DAP requests to all active debug sessions, not just **gdb/gdbtarget** sessions. This breaks debugging with other debug adapters such as **ms-python.debugpy** (Python).

This is a regression — **v2.7.0 does not have this issue**. The problem was introduced in **v2.8.0**.

When a debugpy session is active, **SourceFileHighlighting.handleOnDidChangeActiveTextEditor()** sends a **breakpointLocations** custom request to the Python debug adapter. This causes:

Unhandled: on_breakpointlocations_request not available in PyDevJsonCommandProcessor errors
Breakpoints not being hit — the debug session launches and runs but never stops at breakpoints
Downgrading to v2.7.0 or disabling the CDT GDB extension resolves the issue entirely.

**Root Cause**
I think #224 or https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/537 is where this issue was introduced.

In src/SourceFileHighlighting.ts, the onDidChangeActiveDebugSession handler does not check the session type before storing the session reference and issuing requests:
```typescript
vscode.debug.onDidChangeActiveDebugSession((session) => {
    if (!session) {
        this.clearExecutableLineDecorations(
            vscode.window.visibleTextEditors
        );
    }
    this.activeDebugSession = session;  // ← stores ANY session type
    this.handleOnDidChangeActiveTextEditor(
        vscode.window.activeTextEditor
    );
});
``` 
getBreakpointLocations() then unconditionally sends requests to whatever session is stored:
```typescript
const breakpointLocations = await this.activeDebugSession?.customRequest(
    'breakpointLocations',
    args
);
``` 
**Expected Behavior**
SourceFileHighlighting should only interact with gdb and gdbtarget debug sessions.

**Environment**
VS Code: 1.124.4
CDT GDB Debug Adapter Extension: 2.8.0 (not present in 2.7.0)
ms-python.debugpy: 2026.6.0
OS: Windows

**Steps to Reproduce**

- Install both eclipse-cdt.cdt-gdb-vscode (v2.8.0) and ms-python.debugpy
- Create a Python file with a breakpoint set
- Launch a debugpy debug session
- Observe: session runs to completion without stopping at breakpoints
- Downgrade to v2.7.0 or disable CDT GDB extension → breakpoints work correctly

omarArm:
PR is split in two main commits
1. Fixes https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/232 by filtering debug types where the commands are sent
2. add a command and a setting to enable/disable active source code lines highlighting


src/SourceFileHighlighting.ts
```diff
@@ -28,29 +31,107 @@ export class SourceFileHighlighting {
 
     public activate(): void {
         this.registerToEvents();
+        this.registerCommands();
+        vscode.commands.executeCommand(
+            'setContext',
+            'cdt.debug.sourceCodeHighlightingEnabled',
+            false
```
jreineckearm: I believe we should enable by default to make users aware of this feature. And then give them the ability to opt out.

src/SourceFileHighlighting.ts
```diff
@@ -28,29 +31,107 @@ export class SourceFileHighlighting {
...
+    ): Promise<void> {
+        if (event.affectsConfiguration('cdt.debug.sourceHighlighting')) {
+            this.highlightingEnabled = vscode.workspace
+                .getConfiguration()
+                .get<boolean>('cdt.debug.sourceHighlighting', false);
+            if (!this.highlightingEnabled) {
+                this.clearExecutableLineDecorations(
+                    vscode.window.visibleTextEditors
+                );
+                vscode.commands.executeCommand(
```
jreineckearm: Nit: could pull the context setting out and pass `this.highlightingEnabled` as argument to `executeCommand` to save code lines.

src/SourceFileHighlighting.ts
```diff
@@ -68,6 +149,10 @@ export class SourceFileHighlighting {
         if (!editor) {
             return;
         }
+        if (!this.highlightingEnabled) {
```
jreineckearm: `||` it with the next `if`?

src/SourceFileHighlighting.ts
```diff
@@ -93,6 +178,41 @@ export class SourceFileHighlighting {
         editor.setDecorations(this.executableLineDecorator, decorations);
     }
 
+    private async handleOnDidChangeActiveDebugSession(
+        session: vscode.DebugSession | undefined
+    ): Promise<void> {
+        if (!session) {
+            this.clearExecutableLineDecorations(
```
jreineckearm: Just curious: would this only remove our decorators? Or also those contributed by other extensions?

src/SourceFileHighlighting.ts
```diff
@@ -93,6 +178,41 @@ export class SourceFileHighlighting {
...
+                'setContext',
+                'cdt.debug.sourceCodeHighlightingEnabled',
+                false
+            );
+            this.highlightingEnabled = false;
+            this.activeDebugSession = undefined;
+            return;
+        }
+        if (session.type !== 'gdb' && session.type !== 'gdbtarget') {
+            this.clearExecutableLineDecorations(
```
jreineckearm: I think this is the exact some code block as for `!session`. Maybe we can avoid the duplication?

package.json
```diff
@@ -43,6 +43,15 @@
         ]
       }
     ],
+    "configuration": {
+      "properties": {
+        "cdt.debug.sourceHighlighting": {
+          "type": "boolean",
+          "default": false,
+          "description": "Highlight where breakpoints could be set"
```
jreineckearm: Would make this closer to the shown user commands, i.e. "mark" and "active source lines" and the likes.

src/SourceFileHighlighting.ts
omarArm: I just though that having it on by default might be a bit annoying to users. I will change it to on by default then

src/SourceFileHighlighting.ts
omarArm: supposedly only ours. The implementation of the method `clearExecutableLineDecorations()` has the line `editor.setDecorations(this.executableLineDecorator, []);` which uses the decorator `this.executableLineDecorator` as the intended decorator to be removed

package.json
omarArm: sure, I will change it

src/SourceFileHighlighting.ts
omarArm: done

jreineckearm (approved):
Thanks, changes look good to me. I start liking the line background coloring.
Also tested the settings behavior through the VS Code settings dialog and the newly added command.

We should consider adding some basic documentation. But that should also include the radix commands and the breakpoint modes. Hence, OK to defer to a separate PR.

Merging the changes.

PR #240, opened 2026-07-15
Bump/pin dependencies

jreineckearm:
- Pin all direct production and development dependencies to exact versions for reproducible installs, except `@types/vscode`, which retains its existing compatibility range.
    - Keep `eslint-config-prettier` at 10.1.5 and pin it exactly, avoiding the compromised 10.1.6 and 10.1.7 releases. Version 10.1.8 is the post-incident clean republish without functional changes.
- Update development dependencies in `package.json`:
    - `@eslint/js`: add 10.0.1 as an explicit dependency matching ESLint 10.
    - `@typescript-eslint/eslint-plugin`: 8.35.1 to 8.63.0.
    - `@typescript-eslint/parser`: 8.35.1 to 8.63.0.
    - `@vscode/vsce`: 3.7.1 to 3.9.2.
    - `esbuild-sass-plugin`: 3.3.1 to 3.7.0.
    - `eslint`: 9.30.1 to 10.6.0.
    - `globals`: 16.3.0 to 17.7.0.
    - `prettier`: 3.6.2 to 3.9.5.
        - Apply the resulting formatting change to `src/memory/common/messages.ts`.
    - `typescript-eslint`: 8.35.1 to 8.63.0, restoring compatibility with ESLint 10.
- Patch transitive dependencies through `yarn.lock`:
    - `js-yaml` 3.x: update 3.14.2 to 3.15.0.
    - `js-yaml` 4.x: update 4.1.1 to 4.3.0.
    - `minimatch` 3.x: update 3.1.2 to 3.1.5.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1057,7 +1057,7 @@
     "@types/vscode": "^1.78.0",
     "@typescript-eslint/eslint-plugin": "8.35.1",
     "@typescript-eslint/parser": "8.35.1",
-    "@vscode/vsce": "3.7.1",
+    "@vscode/vsce": "3.9.2",
     "esbuild": "0.28.1",
     "esbuild-sass-plugin": "3.3.1",
     "eslint": "9.30.1",
--- a/package.json
+++ b/package.json
@@ -1065,6 +1065,7 @@
     "eslint-config-prettier": "10.1.5",
     "event-stream": "4.0.1",
     "globals": "17.7.0",
+    "npm-run-all": "4.1.5",
     "prettier": "3.9.5",
     "typescript": "5.8.3",
     "typescript-eslint": "8.63.0"
--- a/package.json
+++ b/package.json
@@ -1051,6 +1051,7 @@
     "react-dom": "19.1.0"
   },
   "devDependencies": {
+    "@eslint/js": "10.0.1",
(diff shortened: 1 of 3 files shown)


omarArm (approved):
LGTM

cwalther:
Out of curiosity (you don’t have to answer if you don’t have time or inclination) – I am a bit out of the loop regarding JavaScript development: What is the rationale for pinning to an exact version in package.json? I always thought the job of package.json was to list the full range of versions that the project is supposed to work with, and narrowing that to exact versions for a reproducible build was the job of yarn.lock / package-lock.json?

jreineckearm:
Correct, that's the idea behind it. That's why we for example didn't catch the corrupted `eslint-config-prettier` version (it was pinned to the last good version in the lockfile).
But there are some triggers which cause the package management tools (npm/yarn) to refresh what they include. Apparently on an initial build without existing lockfile. Or for example when you update the `package.json` entry to a newer minimum version. It automatically checks if something newer than the specified is in range and uses that.
It is widely recommended these days to pin to exact versions so that you don't catch something newer/corrupted by accident. It doesn't help with transient dependencies that request ranges. But reduces the attack surface already.

PR #241, opened 2026-07-15
2.9.0 Release Preparations

jreineckearm:
- Bumps version to 2.9.0
- Bump cdt-gdb-adapter to 1.10.0
- CHANGELOG

PR #242, opened 2026-07-16
Fix version to 2.9.0 for re-publish

jreineckearm:
As the title says, skipping code review to get the fix republished ASAP.
This must have gotten lost when merging in `main` before merging it itself.


cwalther:
Actually no, it wasn’t the merge, it was never part of #241 to begin with. And I didn’t notice because I was focusing on the changelog. Oops!

Thanks for the release!

Issue #244, opened 2026-07-22
cdt.debug.sourceHighlighting written for unrelated debug session


jreineckearm:
The active source code highlighting still seems to be active in non-`gdbtarget`/`gdb` debug environments.
I was debugging something in a Typescript project and at some point I see the
```
    "cdt.debug.sourceHighlighting": false,
```

getting added to the `.vscode/settings.json`.

It might be a side effect of #243 switching to the non-default value `false`. But we need to make sure we better control when this gets written.

jreineckearm:
Gone with fix for #243.

PR #247, opened 2026-07-29
Don't update setting when session is closing

### LINKED ISSUE #243: Mark Active Source Lines is disabled after ending a debug session
### Description

The `cdt.debug.sourceHighlighting` setting is automatically changed to `false` when a GDB debug session ends.

As a result, **Mark Active Source Lines** under the Workspace Settings must be enabled again for every new debug session.

Update, 22/07/2026: In addition to that, I see the setting written to `.vscode/settings.json` for a workspace that doesn't use `gdbtarget/`gdb`. This could be a side effect of this bug. Yet we need to become more robust here.

Related `cdt-gdb-vscode` tickets:
- https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/243
- https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/244

### Steps to Reproduce

1. Make sure the **Mark Active Source Lines** is enabled in Settings.
2. Start a debug session.
4. Confirm that source lines are highlighted.
5. End the debug session.
6. Start another debug session.
7. Notice that source-line highlighting is disabled.

### Expected Behavior

Ending a debug session should not change the persisted `cdt.debug.sourceHighlighting` preference.

omarArm:
when a debug session is shutting down, settings.json should not be updated again. Instead, the latest setting should remain as is

Fixes: https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/243 and https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/244

### DIFF
--- a/src/SourceFileHighlighting.ts
+++ b/src/SourceFileHighlighting.ts
@@ -199,29 +199,20 @@ export class SourceFileHighlighting {
             return;
         }
         this.activeDebugSession = session;
-        await this.handleOnDidChangeActiveTextEditor(
-            vscode.window.activeTextEditor
-        );
+        const shouldEnable = vscode.workspace
+            .getConfiguration()
+            .get<boolean>('cdt.debug.sourceHighlighting', true);
+        if (shouldEnable) {
+            await this.handleEnableSourceFileHighlighting();
+        } else {
+            return;
+        }
     }
 
     private async handleSessionInActive() {
         await this.clearExecutableLineDecorations(
             vscode.window.visibleTextEditors
         );
-        await vscode.commands.executeCommand(
-            'setContext',
-            'cdt.debug.sourceCodeHighlightingEnabled',
-            false
-        );
-        await vscode.workspace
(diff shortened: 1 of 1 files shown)


jreineckearm:
The two issues are fixed with this. But now we are back to seeing `Mark Active Source Lines` in the context menu after session start, even if that is already active as default behavior (setting in settings.json is undefined). Smells a bit like checking somewhere for falsy instead of `false`.

jreineckearm (approved):
Thanks, tested various cases
- switching between multi-core sessions.
- different starting points (entry in settings.json present, false, true)
- using debug connections that are not `gdb` and `gdbtarget`.

No more unexpected writing of `settings.json`, context menu entry carries the correct text (i.e. the highlighted context is not correctly updated). I believe we have this now under control.

Approved, will merge soon. Tempted to make a patch release out of this because it still affects users that don't actively use `gdbtarget`/`gdb` adapters.

